PR #40, opened 2025-12-18
fix: handle error when getting workflow logger

### LINKED ISSUE #39: unhandled error when getting workflow logger in the worker healthcheck job
There's an unhandled error when getting the workflow logger in the worker healthcheck job:
https://github.com/nuonco/nuon/blob/2f1f4a6db92e4dcdeefb71bf6974555b6042453c/services/ctl-api/internal/app/runners/worker/healthcheck_job.go#L36

```
	l, err := log.WorkflowLogger(ctx)
	logStream, err := activities.AwaitCreateLogStreamByOperationID(ctx, req.HealthCheckID)
	if err != nil {
		return &HealthcheckJobRunnerResponse{ShouldRestart: false}, errors.Wrap(err, "unable to create logstream")
	}
```

The err returned from `log.WorkflowLogger(ctx)` is overwritten and never checked.

jgannonjr:
fixes #39


jgannonjr:
sure no problem.  it's rebased now.

PR #55, opened 2025-12-23
chore: remove legacy protobuf references

jonmorehouse:
This PR removes some legacy protobuf references, and removes the need to include the protobuf tool chain at this point.

For context, we used to use protobufs for managing workflow types. However, when we consolidated `workers-executors`
into the `ctl-api` code base, we opted to just use go types all in the same tree.

This PR:

1. removes `pkg/types/workflows`.
1. adds missing types for canarys and infra-tests.
1. removes the need to to include protos in our swag generation.


jonmorehouse:
@nuonco/team please take a look.

onprem (dismissed):

onprem (approved):

PR #56, opened 2025-12-23
feat: dynamic comments for isntall config generation

someshkoli:
fork branch pr for go-toml: https://github.com/someshkoli/go-toml/pull/1/changes

### DIFF
--- a/bins/cli/internal/services/installs/app_install_syncer.go
+++ b/bins/cli/internal/services/installs/app_install_syncer.go
@@ -16,8 +16,10 @@ import (
 	"github.com/nuonco/nuon/pkg/generics"
 )
 
-const ManagedByNuonCLIConfig = "nuon/cli/install-config"
-const ManagedByNuonDashboard = "nuon/dashboard"
+const (
+	ManagedByNuonCLIConfig = "nuon/cli/install-config"
+	ManagedByNuonDashboard = "nuon/dashboard"
+)
 
 const defaultPollDuration = time.Second * 10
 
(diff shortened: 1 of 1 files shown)


services/ctl-api/internal/app/installs/service/generate_cli_install_config.go
```diff
@@ -107,24 +106,21 @@ func (s *service) genCLIInstallConfig(ctx context.Context, installID string) (*c
...
+			// currently group name is being dropped and is not being passed forward
+			inputGroupInputs["__nuon.input.group"] = groupName
+			installCfg.InputGroups = append(installCfg.InputGroups, inputGroupInputs)
 		}
 	}
 
-	slices.Sort(keys)
-	for _, k := range keys {
-		installCfg.InputGroups = append(installCfg.InputGroups, inputGroups[k])
-	}
```
onprem, line 127: this was added as we need to generate the same config with same set of inputs. Since the toml lib at the time didn't guarantee preserving order in maps (or something similar, IIRC, like go range over map not guaranteeing  the order). Will the order be preserved without this in the new implementation?

someshkoli: interesting, i think it said something about ordering, but leme just revert this to avoid any sideeffects

someshkoli: yeah I though it was redundant, move em back in also wrote ut to make sure its workign correctly

PR #58, opened 2025-12-23
fix: handle temporal workflow not found during cancel

erickyellott:
When canceling a workflow, sometimes the underlying temporal workflow has already gone away. In this case, we just swallow the error.


fidiego (approved):
LGTM

PR #62, opened 2025-12-25
fix: add language server docs

sharkymark:
The language server was only mentioned in a summary changelog so am adding a section on Configuration Files to educate users creating apps to try it out.

It is beta and part of Nuon Labs so clearly designated it as beta and to report issues on GitHub.


jonmorehouse (approved):
Looks good! Do you mind adding some details in the PR description, for posterity?

PR #66, opened 2025-12-28
feat: add new temporal-gen generator

jonmorehouse:
We have relied on a wrapper interface for generating methods to interact with native go code via temporal. 

This legacy library is no longer mainted, relies on a third-party library we do not want to build on, and is missing 
critical functionality for the new queues. This PR introduces a rewrite, where we introduce a new implementation with 
many new benefits:

1. First, it's fully tested. We have a test suite to verify all of the generated code wrappers and annotations.
1. It supports generating workflow activity wrappers as well as activity wrappers for plan functions. This means we can 
   take a plain go struct and with _no_ changes make it temporal compatible.
1. It supports methods for signals, queries and updates.
1. It supports variadic options, so you can use a generated wrapper method, but also send custom parameters in.

Further, it fixes some issues internally such as relying on internal methods to create workflow-ids and others.

We plan to roll this out across the entire repo in the next month or so.

### DIFF
--- a/pkg/gen/temporal-gen-v2/examples/activity.go
+++ b/pkg/gen/temporal-gen-v2/examples/activity.go
@@ -1,6 +1,6 @@
 package examples
 
-//go:generate go run ../main.go .
+//go:generate go run ../main.go generate .
 
 import (
 	"context"


jonmorehouse:
It's worth mentioning that this was originally opened as a PR pre OSS, and is not going to be used in this first commit. We will slowly roll this out procedurally.

sharkymark (approved):

PR #82, opened 2026-01-05
feat: add latest build to app component

nnnnat:
- Added latest_build field to Component model JSON response
- Updated `/v1/apps/:app-id/components` endpoint to preload ComponentConfigs and ComponentBuilds
- Modified Component AfterQuery hook to populate LatestBuild from preloaded data
- Updated `ComponentsTable.tsx` in UI to use the latest build status

### DIFF
--- a/services/ctl-api/internal/app/components/service/get_app_components.go
+++ b/services/ctl-api/internal/app/components/service/get_app_components.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/nuonco/nuon/services/ctl-api/internal/app"
 	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/db"
-	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/db/plugins/views"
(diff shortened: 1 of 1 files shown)


services/ctl-api/internal/app/components/service/get_app_components.go
```diff
@@ -72,7 +74,14 @@ func (s *service) getAppComponents(ctx *gin.Context, appID, q string, types []st
 	tx := s.db.WithContext(ctx).
 		Scopes(scopes.WithOffsetPagination).
 		Where("id IN ?", []string(appCfg.ComponentIDs)).
-		Preload("Dependencies")
+		Preload("Dependencies").
+		Preload("ComponentConfigs", func(db *gorm.DB) *gorm.DB {
+			return db.Order(views.TableOrViewName(s.db, &app.ComponentConfigConnection{}, ".created_at DESC"))
```
ALPHAvibe: This will preload all ComponentConfigConnections filtered by matching Component.IDs from the query results for components. This is potentially an unbounded request that can OOM the api.

PR #85, opened 2026-01-05
fix: Improved error reporting by including source file information for configuration parsing and validation errors.

RealHarshThakur:
Small QoL improvements for better source file information and improved error handling- 
For example:
* Commenting out a component file used to error out. We wouldn't know which file caused this. 
* Commenting out a file now gets ignored, so it doesn't cause an error on apps sync. 
* We had error messages with component indexes like this:

> " components.10: Invalid type. Expected: object, given: null components.9: Invalid type. Expected: object, given: null ". 

Now we see it with the file name:

> `components/external_secrets.toml: error decoding config 
>  error decoding '': must set `type` on each component`

### DIFF
--- a/pkg/config/validate/json.go
+++ b/pkg/config/validate/json.go
@@ -2,12 +2,19 @@ package validate
 
 import (
 	"context"
+	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
+
+	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/nuonco/nuon/pkg/config"
 	"github.com/nuonco/nuon/pkg/config/schema"
-	"github.com/nuonco/nuon/pkg/generics"
 )
+
+// componentIndexPattern matches error fields like "components.9" or "components.9.name"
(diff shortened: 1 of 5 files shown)


stephennuon (approved):
LGTM 🚀

pkg/config/validate/json.go
```diff
@@ -18,7 +25,44 @@ func ValidateJSONSchema(ctx context.Context, c *config.AppConfig) error {
...
+
+	for _, err := range errs {
+		field := err.Field()
+		desc := err.Description()
+
+		// Try to replace component index with file name
+		if matches := componentIndexPattern.FindStringSubmatch(field); len(matches) == 2 {
+			if idx, parseErr := strconv.Atoi(matches[1]); parseErr == nil {
+				if idx >= 0 && idx < len(c.Components) && c.Components[idx] != nil {
+					sourceFile := c.Components[idx].GetSourceFile()
```
someshkoli, line 50: c.Components[idx].Sourcefile ?

PR #102, opened 2026-01-08
feat: Add OPA engine Policy API & workflows

RealHarshThakur:
This change introduces support for Open Policy Agent (OPA) as a policy engine, alongside the existing Kyverno engine, and refactors policy validation and execution.

### DIFF
--- a/services/dashboard-ui/src/providers/log-stream-provider.tsx
+++ b/services/dashboard-ui/src/providers/log-stream-provider.tsx
@@ -33,13 +33,13 @@ export function LogStreamProvider({
     stopPolling,
   } = usePolling<TLogStream>({
     initData: initLogStream,
-    path: `/api/orgs/${org.id}/log-streams/${initLogStream.id}`,
+    path: `/api/orgs/${org.id}/log-streams/${initLogStream?.id}`,
     pollInterval,
     shouldPoll,
   })
 
   useEffect(() => {
-    if (!logStream.open) {
+    if (!logStream?.open) {
       stopPolling()
     }
   }, [logStream])


services/ctl-api/internal/pkg/flow/execute_workflow_step.go
```diff
@@ -642,3 +715,51 @@ func (c *WorkflowConductor[DomainSignal]) handlePlanOnlyApproval(ctx workflow.Co
...
+	// so we can use generated activity options instead of manually specifying them here.
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout:    1*time.Minute + 30*time.Second,
+		ScheduleToCloseTimeout: 2 * time.Minute,
+		RetryPolicy:            &temporal.RetryPolicy{MaximumAttempts: 1},
+	}
+	policyCtx := workflow.WithActivityOptions(ctx, ao)
+
+	var futures []workflow.Future
+	for _, policy := range prepResult.Policies {
```
onprem, line 770: with too many policies, this can result in history too large errors. Maybe this should be done in it's own child workflow, with continue as new batching.

services/ctl-api/internal/pkg/flow/execute_workflow_step.go
```diff
@@ -171,6 +172,78 @@ func (c *WorkflowConductor[DomainSignal]) executeFlowStep(ctx workflow.Context,
...
+			}
+			return false, fmt.Errorf("policy violations found: %d deny violations", len(denyViolations))
+		}
+
+		if len(warnViolations) > 0 {
+			if updateErr := statusactivities.AwaitPkgStatusUpdateFlowStepStatus(ctx, statusactivities.UpdateStatusRequest{
+				ID: step.ID,
+				Status: app.CompositeStatus{
+					Status:   step.Status.Status,
+					Metadata: map[string]any{"policy_violations": violations},
```
onprem: small nit, maybe we should make it explicit that these are warnings in the metadata key.

Also for the value `violations` is indeed only warnings as we exit early if there are any denyViolations, but maybe it'll be nice to be explicitly use `warnViolations` here.

```suggestion
					Metadata: map[string]any{"policy_violation_warnings": warnViolations},
```

services/ctl-api/internal/pkg/flow/execute_workflow_step.go
```diff
@@ -642,3 +715,51 @@ func (c *WorkflowConductor[DomainSignal]) handlePlanOnlyApproval(ctx workflow.Co
...
+	var allViolations []activities.PolicyViolation
+	for _, fut := range futures {
+		var result activities.EvaluateSinglePolicyResult
+		if err := fut.Get(ctx, &result); err != nil {
+			return nil, errors.Wrap(err, "policy evaluation failed")
+		}
+		allViolations = append(allViolations, result.Violations...)
+	}
+
+	return allViolations, nil
```
onprem, line 789: small nit again,
maybe we can move the logic to warnings vs deny violations in this helper itself and return two arrays, one with warnings, and one with deny violations directly from here?

https://github.com/nuonco/nuon/pull/102/changes/e111f8cbbd08064186c0ebd23838432223fba49c#diff-09833bd453ea4317d4d5e8a435c85f953cc400d2d6493914c5f504337bbf003cR193-R201

This will make the execute step function a bit cleaner.

services/ctl-api/internal/pkg/workflows/workflow/activities/evaluate_single_policy.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type EvaluateSinglePolicyRequest struct {
+	PolicyID  string `json:"policy_id" validate:"required"`
+	Contents  string `json:"contents" validate:"required"`
+	InputJSON []byte `json:"input_json" validate:"required"`
+}
+
+type EvaluateSinglePolicyResult struct {
+	Violations []PolicyViolation `json:"violations" temporaljson:"violations,omitempty"`
```
onprem, line 21: Similarly here we can return warn violations and deny violations separately. As we anyways have then separate internally. Also makes it easier for consumer.

services/ctl-api/internal/pkg/workflows/workflow/activities/prep_policy_evaluation.go
```diff
@@ -0,0 +1,300 @@
...
+		IsSandbox:     true,
+	}, nil
+}
+
+func (a *Activities) getPoliciesConfigByAppConfigID(ctx context.Context, appConfigID string) (*app.AppPoliciesConfig, error) {
+	var policiesConfig app.AppPoliciesConfig
+	res := a.db.WithContext(ctx).
+		Where("app_config_id = ?", appConfigID).
+		Preload("Policies").
+		First(&policiesConfig)
```
onprem, line 174: Can there be multiple AppPoliciesConfig with same `app_config_id`? In that case maybe we also need to add `.OrderBy("created_at DESC")` to get the correct (latest) one?

RealHarshThakur: great catch.

services/ctl-api/internal/pkg/flow/execute_workflow_step.go
RealHarshThakur: Broke down the execute flow step function, lot cleaner now.

pkg/config/validate/validate_policies_test.go
```diff
@@ -13,18 +12,89 @@ func TestValidatePolicyType(t *testing.T) {
...
+		{config.AppPolicyTypeTerraformModule, config.AppPolicyEngineOPA, false},
+		{config.AppPolicyTypeHelmChart, config.AppPolicyEngineKyverno, true},
+		{config.AppPolicyTypeHelmChart, config.AppPolicyEngineOPA, false},
+		{config.AppPolicyTypeKubernetesManifest, config.AppPolicyEngineKyverno, true},
+		{config.AppPolicyTypeKubernetesManifest, config.AppPolicyEngineOPA, false},
+		// empty engine skips check
+		{config.AppPolicyTypeKubernetesCluster, "", false},
+	}
+
+	for _, test := range tests {
```
jonmorehouse: I've found it a bit nicer to use a map[string]<type> here, so you have a mapping of tests to a name, and then add them as a sub-test underneath. It makes discovery much better later one, when debugging something.

services/ctl-api/internal/pkg/workflows/workflow/activities/evaluate_single_policy.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// @temporal-gen activity
```
jonmorehouse, line 24: If you have the appetite, you can use `temporal-gen-v2`. Maybe in a follow up? 

TLDR, you don't have to define the request type, and the signature of the private method can be used to generate the public activity definition.

services/ctl-api/internal/pkg/workflows/workflow/activities/evaluate_single_policy.go
```diff
@@ -0,0 +1,129 @@
...
+// @max-retries 1
+// @schedule-to-close-timeout 2m
+// @start-to-close-timeout 1m30s
+func (a *Activities) EvaluateSinglePolicy(ctx context.Context, req *EvaluateSinglePolicyRequest) (*EvaluateSinglePolicyResult, error) {
+	l := temporalzap.GetActivityLogger(ctx)
+	l = l.With(zap.String("policy_id", req.PolicyID))
+
+	l.Info("evaluating policy")
+
+	var input interface{}
```
jonmorehouse: prefer `any`

services/ctl-api/internal/pkg/workflows/workflow/activities/prep_policy_evaluation.go
```diff
@@ -0,0 +1,301 @@
...
+	Severity string `json:"severity" temporaljson:"severity,omitempty"` // "deny" or "warn"
+}
+
+type PolicyToEvaluate struct {
+	PolicyID  string `json:"policy_id" temporaljson:"policy_id,omitempty"`
+	Contents  string `json:"contents" temporaljson:"contents,omitempty"`
+	InputJSON []byte `json:"input_json" temporaljson:"input_json,omitempty"`
+}
+
+type PrepPolicyEvaluationResult struct {
```
jonmorehouse, line 35: FYI, it seems that quite a bit is happening in here. Is this idempotent?

I can't help but feel that long term, evaluating a policy should be a child-workflow. I see why it shouldn't be now, but just thinking about a future where this grows in scope.

PR #127, opened 2026-01-10
fix: short circuit action crons if install not provisioned

erickyellott:
Action cron jobs were queueing up for installs that hadn't finished provisioning (waiting on the install stack to be run by the user). This was causing the temporal workflow history size to get too large after several weeks. 

Here we skip those crons if the stack is known to be provisioning/deprovisioning.


fidiego (approved):
🙏

PR #134, opened 2026-01-12
chore: check if generated code is commited

RealHarshThakur:
TLDR: Checks if generated code or docs are committed. 
We use code & doc generators heavily, some of which is committed to the repo. 
However, I've noticed that sometimes it can be quite easy to make a quick change in ctl-api and forget to update the SDK. 

This workflow ensures that all necessary changes are committed by :
* Using the script for reset-generated-code that triggers all the generators. 
* we then use git to check if any of the tracked files are committed or not.

### DIFF
--- a/bins/runner/internal/jobs/sync/oci/handler.go
+++ b/bins/runner/internal/jobs/sync/oci/handler.go
@@ -34,18 +34,20 @@ var _ jobs.JobHandler = (*handler)(nil)
 type HandlerParams struct {
 	fx.In
 
-	V         *validator.Validate
-	APIClient nuonrunner.Client
-	Config    *internal.Config
-	OCICopy   ocicopy.Copier
+	V           *validator.Validate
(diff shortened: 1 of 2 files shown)


jonmorehouse:
@RealHarshThakur you mind updating this description to share more about the workflow here?

.github/workflows/check-generated.yml
```diff
@@ -0,0 +1,56 @@
...
+          cp -r scripts/* bins/nuonctl/scripts/
+          chmod +x bins/nuonctl/scripts/*
+
+      - name: Generate code
+        run: |
+          CI=true nuonctl scripts reset-generated-code
+
+      - name: Check for uncommitted changes
+        run: |
+          if ! git diff --exit-code -- sdks/ docs/; then
```
erickyellott, line 50: Don't we also have generated temporal code as well?

erickyellott (approved):

PR #137, opened 2026-01-12
chore(ctl-api): patch endpoint for global IdP configs

fidiego:
- **chore(ctl-api): patch endpoint for global idps**
- **chore(ctl-api): simplify**

### DIFF
--- a/services/dashboard-ui/src/middleware.ts
+++ b/services/dashboard-ui/src/middleware.ts
@@ -1,16 +1,29 @@
 import { NextResponse, type NextRequest } from 'next/server'
-import { USE_AUTH_SERVICE } from '@/configs/auth'
+import { USE_AUTH_SERVICE, AUTH_SERVICE_URL, APP_URL } from '@/configs/auth'
 import { auth0 } from '@/lib/auth'
+import { getSession } from '@/lib/auth-server'
 
 export default async function middleware(request: NextRequest) {
   const { pathname } = new URL(request.url)
 
   if (USE_AUTH_SERVICE) {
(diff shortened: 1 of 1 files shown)


services/ctl-api/internal/app/auth/service/admin_patch_identity_provider.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+
+	// Load existing provider
+	var ip app.IdentityProvider
+	result := s.db.WithContext(ctx).Where("id = ?", identityProviderID).First(&ip)
+	if result.Error != nil {
+		ctx.Error(fmt.Errorf("identity provider not found: %w", result.Error))
+		return
+	}
```
erickyellott, line 71: Will these throw 500 level errors? do we need to pull in the 4xx level `stderr.ErrUser` error type for these?

erickyellott (approved):

services/ctl-api/internal/app/auth/service/admin_patch_identity_provider.go
fidiego: good call.

fidiego: this returns a 404. 
image

erickyellott: Dang, bad UI. Had selected a block. Was more worried about these that aren't error wrapped with `%w`:

> ctx.Error(fmt.Errorf("identity_provider_id is required"))

PR #138, opened 2026-01-12
feat(ui): redirect to auth service signin if session is null

nnnnat:
Handle sign-in redirect if expired auth service token in nextjs middleware


stephennuon (approved):
🚀

PR #140, opened 2026-01-12
feat(auth):  styles for auth service pages

nnnnat:
Simple styling of auth service pages using tailwindcss cdn and assets from the nuon.co website
auth-styles

### DIFF
--- a/services/ctl-api/internal/app/auth/service/templates/base.html.tmpl
+++ b/services/ctl-api/internal/app/auth/service/templates/base.html.tmpl
@@ -1,18 +1,52 @@
-<!-- vi: set syntax=html -->
-{{ define "auth/head.tmpl" }}
+{{define "head.tmpl"}}
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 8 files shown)


services/ctl-api/internal/app/auth/service/templates/base.html.tmpl
```diff
@@ -1,18 +1,52 @@
...
+{{define "head.tmpl"}}
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta http-equiv="X-UA-Compatible" content="ie=edge">
-{{ end }}
+<link rel="icon" type="image/svg+xml" href="https://nuon.co/favicon.svg">
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap" rel="stylesheet">
+<script src="https://cdn.tailwindcss.com"></script>
```
fidiego, line 8: i think we can grab the tailwindcli and use it to grab only what we need to avoid including everything. 
i'll put that in an `nctl` script for your consideration.

PR #155, opened 2026-01-14
test(ui): adds playwright setup

nnnnat:
- Adds a playwright e2e testing framework
- `Adds` a process for capturing user flows in markdown so claude can write the playwright test
- Adds a simple smoke test and create install flow test


nnnnat:
Implemented in a different PR

PR #159, opened 2026-01-15
fix: uncomment enable audit param

ALPHAvibe:
This was causing a nil pointer error when we re-enabled our endpoint audits.


fidiego (approved):
LGTM

PR #171, opened 2026-01-17
chore: regenerate sdks

erickyellott:
The codegen files have drifted with recent changes. Here we get them in sync. 

We need to add a PR check that checks for this in a followup.

Edit: Harsh is already ahead of me: https://github.com/nuonco/nuon/pull/134


RealHarshThakur:
@erickyellott closing this as I merged that PR & have regen the code.

PR #182, opened 2026-01-19
feat: remove password authentication from Azure runner VMs 

jordan-acosta:
Remove hard-coded password authentication from runner VM (was only needed for debugging.) Because Azure requires either SSH or user/pass, add throw-away SSH key. We do not retain a private key, to this should lock down the VM in the same way we lock down the AWS runner VM.


sharkymark (approved):

fidiego (approved):
LGTM 

wondering if it'd be possible to include something like that conditionally. would be great for dev/debugging.

PR #184, opened 2026-01-20
fix: fetch workflow step again before starting  exeuction block

someshkoli:


services/ctl-api/internal/pkg/flow/execute_workflow_step.go
```diff
@@ -169,14 +171,15 @@ func (c *WorkflowConductor[DomainSignal]) executeFlowStep(ctx workflow.Context,
...
+				return refetchStepsInfo, errors.Wrap(err, "unable to mark step as error")
 			}
 
-			return false, errors.Wrap(err, "failed to handle noop plan")
+			return refetchStepsInfo, errors.Wrap(err, "failed to handle noop plan")
 		}
 
 		if !flw.PlanOnly {
-			return true, nil
+			refetchStepsInfo = true
```
someshkoli, line 181: @RealHarshThakur here we set it true

services/ctl-api/internal/pkg/flow/execute_workflow_step.go
```diff
@@ -270,10 +275,10 @@ func (c *WorkflowConductor[DomainSignal]) executeFlowStep(ctx workflow.Context,
 				},
 			},
 		}); err != nil {
-			return false, errors.Wrap(err, "unable to update step to success status")
+			return refetchStepsInfo, errors.Wrap(err, "unable to update step to success status")
```
someshkoli, line 278: here we return same value

PR #189, opened 2026-01-20
fix(ui): sse log connection

nnnnat:
- keep sse log connection open until user closes logs
- close top most panel using esc key


stephennuon (approved):
LGTM

PR #190, opened 2026-01-21
fix: add cleanup for generated config files in tests

someshkoli:
Added cleanup step to remove generated config files after test.


pkg/config/generator/generate_test.go
```diff
@@ -274,4 +274,10 @@ func TestGenerateWithInstanceValues(t *testing.T) {
 	}
 
 	t.Logf("Successfully verified instance values in generated TOML files")
+
+	// cleanup
+	err = os.RemoveAll("./test-config-init")
```
RealHarshThakur: can we do this as a defer?

RealHarshThakur (dismissed):

RealHarshThakur (approved):

PR #206, opened 2026-01-22
fix: middleware fx registration

ALPHAvibe:


services/ctl-api/internal/pkg/api/base.go
```diff
@@ -56,18 +56,33 @@ func (a *API) registerMiddlewares() error {
...
+
+	// validate no duplicates in configured middlewares
+	seenConfigured := make(map[string]bool)
+	for i, middleware := range a.configuredMiddlewares {
+		if seenConfigured[middleware] {
+			return fmt.Errorf("duplicate middleware in configuration for %s API: %s", a.name, middleware)
+		}
+		seenConfigured[middleware] = true
+
+		a.l.Info(fmt.Sprintf("rb - %s API [%d/%d] registering middleware: %s", a.name, i+1, len(a.configuredMiddlewares), middleware))
```
ALPHAvibe, line 77: thesee are just for troubleshooting and needs to be removed before merging

PR #707, opened 2026-03-26
fix: remove view usage for app-config

ALPHAvibe:
Database CPU is elevated. app_configs_view_v2 is the #1 CPU consumer (0.72 AAS, ~6 calls/sec), computing ROW_NUMBER() on every read just to derive a version number. Several other queries are missing indexes on their primary filter columns.

  Changes

  Disable app_configs_view_v2 — top CPU contributor:
  - UseView() returns false — queries now hit the base app_configs table directly
  - version is now a stored column computed at write time (BeforeCreate: COUNT(*) + 1)
  - Backfill migration (095) populates existing rows with the same values the view produced
  - Removed the v2 view from Views() to prevent CREATE OR REPLACE conflicts
  - Updated component_config_connections_v1.sql to use the stored version column from app_configs instead of recomputing it

  Missing indexes:
  - install_workflow_steps — (install_workflow_id, deleted_at)
  - app_awsiam_policy_configs — (app_awsiam_role_config_id, deleted_at)
  - component_config_connections — (app_config_id, deleted_at)
  - public_git_vcs_configs — (component_config_id, component_config_type, deleted_at)
  - install_inputs — (install_id, deleted_at)


stephennuon (approved):

stephennuon (dismissed):

services/ctl-api/internal/app/app_config.go
erickyellott, line 1: * Should we add a unique index on (app_id, version)?
* When computing the version, do we need to include deleted rows explicitly? (I think gorm filters them out automatically)
* Should we increment using count, or selecting the last created or max? This is probably minor.

services/ctl-api/internal/app/app_config.go
```diff
@@ -87,8 +87,8 @@ type AppConfig struct {
 	// individual pointers
 	InstallAWSCloudFormationStackVersion []InstallStackVersion `json:"-" gorm:"constraint:OnDelete:CASCADE;" temporaljson:"install_aws_cloud_formation_stack_version,omitzero,omitempty"`
 
-	// fields that are filled in via after query or views
-	Version int `json:"version,omitzero" gorm:"->;-:migration" temporaljson:"version,omitzero,omitempty"`
+	// version is computed at write time as the sequential config number for this app
+	Version int `json:"version,omitzero" gorm:"default:0" temporaljson:"version,omitzero,omitempty"`
```
erickyellott: Current app configs start with 1, not 0. Should we default to 1 instead of 0?

Looks like BeforeCreate will properly increment this by 1 before save, but I can see this being confusing.

services/ctl-api/internal/app/app_config.go
ALPHAvibe: honestly i dont think we should recompute version once set
unique might endup making the change bigger,  cause will have to make the property nil/null to add the property.
last max is always more optimal but needs index

erickyellott: > honestly i dont think we should recompute version once set

Agreed, but not sure if we're accounting for deleted rows when setting it via:

`tx.Model(&AppConfig{}).Where("app_id = ?", a.AppID).Count(&count)`

Think this is the case:

> You'd need `Unscoped()` to include soft-deleted rows.

erickyellott: > unique might endup making the change bigger, cause will have to make the property nil/null to add the property.

Ah good, point, prob need to roll this out in 2 phases, to remove nullable. We could add the unique constraint then. 

Mostly want to avoid any race conditions, accidents where we end up with 2 latest configs somehow.

ALPHAvibe: good catch just need to add unscoped

services/ctl-api/internal/app/app_config.go
ALPHAvibe: i can just not add default right now honestly 0 is kinda already implied, and the before create should avoid setting 0.

services/ctl-api/internal/pkg/db/viewsql/component_config_connections_v1.sql
erickyellott, line 33: can we just replace `app_configs_for_component_configs` with `app_configs` on this line? and drop the CTE?

erickyellott (dismissed):

services/ctl-api/internal/pkg/db/viewsql/component_config_connections_v1.sql
ALPHAvibe: this actually won't be used anymore, but i probably shouldn't change the view. going to revert and leave a comment

erickyellott (approved):

erickyellott (dismissed):

PR #799, opened 2026-04-03
feat: runner iid auth

erickyellott:
Adds AWS EC2 Instance Identity Document (IID) as a new runner
authentication method, alongside the existing STS presigned URL flow.

### Changes

**Runner binary** — new `bins/runner/internal/pkg/aws/iid.go` fetches
the IID document and RSA-2048 PKCS7 signature from IMDSv2. The
`fetch-token` command and token-refresh loop both respect
`RUNNER_AUTH_METHOD=iid`, defaulting to `sts` so existing runners are
unaffected.

**API (`POST /v1/runner-auth/aws-iid`)** — new public endpoint that:
1. Verifies the PKCS7 signature against bundled AWS regional certificates
2. Confirms the IID's `accountId` matches the install's stack outputs
3. Issues a runner token on success


services/ctl-api/internal/app/runner-auth/service/aws_certs.go
erickyellott, line 1: @fidiego Hella certs are scraped from the AWS docs (PITA) for each region and put in here. Is there a better way to store these? Do we need to worry about them rotating at some point?

RealHarshThakur: can we download them at startup ?

erickyellott: I couldn't find an endpoint or repo with them easily digestible. I had to bully Claude into scraping them.

RealHarshThakur: I think in the worst case, we can't do it at startup, I think we should make it a config file rather than hardcoding. 
So in case, we have to update, it would be just a config change

services/ctl-api/internal/app/runner-auth/service/runner_auth_aws_iid.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/nuonco/nuon/services/ctl-api/internal/app"
+	"github.com/nuonco/nuon/services/ctl-api/internal/middlewares/stderr"
+)
+
+var awsAccountIDPattern = regexp.MustCompile(`^\d{12}$`)
+
+// InstanceIdentityDocument represents the JSON document returned by
+// the EC2 IMDS at /latest/dynamic/instance-identity/document.
+type InstanceIdentityDocument struct {
```
jonmorehouse: In a follow up, can we also move this functionality for verifying the document into `pkg/runner/auth`. 

cc @im-Amitto and @RealHarshThakur, so we can do so for azure and GCP

RealHarshThakur (approved):

PR #825, opened 2026-04-07
feat: add label support for common objects

jonmorehouse:
This PR adds label support for common objects in the platform:

* installs - user manageable via the ui or config-file.
* runners - internally managed
* components - user managed via the ui or config-file.
* actions - user managed via the ui or config-file.


nnnnat (dismissed):

stephennuon (approved):
LGTM

PR #859, opened 2026-04-09
fix: V2 provision signal supports GCP

jordan-acosta:
Generating install stack version wasn't working for GPC apps when using the new queue system. We added the new V2 signal used by the queues before starting on GCP support, so the new code path wasn't handling the `gcp` runner type.

This updates the V2 signal code path to handle the `gcp` runner type like the old code path does.


jordan-acosta:
Tested locally and confirmed working.

stephennuon (approved):

PR #863, opened 2026-04-09
feat: (installs) Add per-install stack template overrides

onprem:
Allow vendors to override VPC, Runner, and custom nested stack templates on a per-install basis via InstallConfig.

- Add `VPCNestedTemplateURL`, `RunnerNestedTemplateURL`, and `CustomNestedStacks` fields to `InstallConfig` model
- Expose new fields in Create/Update `InstallConfig` API endpoints (with validation)
- Use `InstallConfig` to apply install-level overrides before template rendering in `GenerateInstallStackVersion` workflow

The overrides are applied such that:
- The values defined in the App Config are used as defaults, and in case `VPCNestedTemplateURL` or `RunnerNestedTemplateURL` in `InstallConfig` is non-empty, we override. Vendors don't need to define both, they can only override one of these if they wish to, and the other stack template url will be picked from the app's config.
- For `CustomNestedStacks`: same-name stacks are replaced, while stacks new names appended to the list from App's config.


fidiego (dismissed):
LGTM

fidiego (approved):
LGTM

PR #871, opened 2026-04-09
fix(ctl-api): create fake gcp stack outputs in sandbox mode

jordan-acosta:
## Problem

We were just creating fake AWS outputs regardless of runner type. This is an issue testing with GCP now we're trying to use the stack outputs.

## Solution

If the runner type is `gcp`, create fake outputs modeled on what our GCP install stack creates.


jordan-acosta:
Tested locally with gke-simple.

stephennuon (approved):

PR #875, opened 2026-04-10
chore: check in generated code

RealHarshThakur:
# Check in all generated code, remove CI generation steps

## The problem

This monorepo generates a lot of code. Temporal activity and workflow wrappers, mocks for test interfaces, Swagger/OpenAPI specs for all three API surfaces (public, admin, runner), SDK clients in `nuon-go` and `nuon-runner-go`, and API reference documentation. Roughly 500 files across the repo are machine-generated.

None of this is optional. If you add a Temporal activity, change a Go interface, or modify an API endpoint, you must run code generation before you can even compile and test locally. Generated code is a prerequisite for creating a working PR, not an afterthought.

Despite this, we treat generated code inconsistently today:
- `sdks/` and `docs/` are checked into git, with a `check-generated` CI workflow that regenerates and diffs them
- `_gen.go` files, mocks, and swagger docs (the other ~500 files) are gitignored and regenerated inside Docker builds, invisible in PRs, never reviewed

The `check-generated` workflow was meant to be the safety net for the first group. In practice, it has been unreliable: flaky failures, slow runs pulling in the full generation toolchain (`swag`, `mockgen`, `temporal-gen-v2`, `templ`), and false negatives that let stale code through. When it breaks, the response is usually to re-run or ignore it.

### Why just removing the check is not enough

If we remove the CI check without any replacement, we lose even the partial protection it provided. The CLI imports `sdks/nuon-go`. The dashboard-ui generates its TypeScript API types from the swagger spec. The published API reference docs come from the same generated swagger output. If any of these go stale after an API change, nothing catches it:

- A new response field added in `ctl-api` silently disappears stays missing from Go SDK
- A new endpoint never shows up in the API docs or the dashboard's generated types
- Swagger specs drift from the actual API behavior with no visible failure

All of which causes surprise & sometimes pain for the next developer who finds these changes. 
Developers already have to generate code to build locally. The question is whether we also track the output. Without tracking it, there is no enforcement at all beyond hoping people remember.

## What this PR does

Make the implicit requirement explicit: since developers must generate code to build locally, that output should be committed. CI then builds and tests from committed source only, with no generation step anywhere.

If you change an interface but forget to regenerate, the build fails. No separate workflow needed.

### Changes

| File | Change |
|---|---|
| `.github/workflows/check-generated.yml` | Deleted. |
| `.gitignore` | Removed `*_gen.go`, `*mock*.go`, swagger docs, and other generated file patterns. |
| `.gitattributes` | Added `linguist-generated=true` for all generated file paths (collapses diffs in PRs). |
| `services/ctl-api/Dockerfile` | Removed `generated` stage and all `COPY --from=generated` lines. Copies full `docs/` directory instead of only description files. |
| `pkg/Dockerfile` | Removed `go generate ./pkg/...` step. |
| `images/go-base/Dockerfile.build` | Removed `go generate` from `pkg-base` stage. |

### Files to commit

Roughly 500 previously-gitignored generated files need to be added:
- Temporal activity/workflow wrappers (`*_gen.go`) across `services/ctl-api/` and `pkg/`
- Generated mocks (`*mock*.go`) across `pkg/`
- Swagger/OpenAPI specs (`services/ctl-api/docs/{public,admin,runner}/`)

### Developers already generate before every PR

You cannot build `ctl-api` without the Temporal `_gen.go` files. You cannot run tests without the mocks. You cannot compile the service without the swagger `docs.go`. Developers are already running generation as part of their workflow. Committing the output is just tracking what they already produce.

### Build failures replace the flaky CI check

With generated code committed, the compiler becomes the enforcement mechanism:
- A missing `_gen.go` file fails the build (unresolved import)
- A stale mock with a wrong method signature fails the build (type error)
- A missing swagger `docs.go` fails the build (import used by `main.go`)

This is more reliable than a CI workflow that regenerates from scratch and diffs.

### One model instead of two

Today we have checked-in SDKs verified by a flaky workflow, and gitignored service code regenerated in Docker. Two sets of rules, two failure modes, two mental models. After this change, everything is committed, everything is visible, everything is enforced by the build.

### Faster Docker builds

The `ctl-api` Dockerfile had a dedicated `generated` stage that ran `go generate ./...` and copied the output into the `build`, `unit-test`, and `lint` stages. This required the full generation toolchain in the build image. That stage is now gone.

### The Go ecosystem convention

The official Go blog post on `go generate`:

> "`go generate` is not part of `go build`. It is intended to be used by the author of the Go package, not its clients."
>
> "once the file is generated (and tested!) it must be checked into the source code repository to be available to clients"
>
> https://go.dev/blog/generate

Kubernetes, grpc-gateway, and templ all follow this pattern: developers generate locally and commit the output.

### What build enforcement catches vs. what it doesn't

| Scenario | Caught? | How |
|---|---|---|
| Missing generated file | Yes | Build fails, import can't resolve |
| Incompatible change (removed/renamed field) | Yes | Build fails, type mismatch |
| Stale mock (interface method changed) | Yes | Build fails, signature mismatch |
| Stale swagger docs (new endpoint added) | No | Swagger JSON compiles fine, just missing the new endpoint |
| Stale SDK (new field added to response) | No | CLI compiles fine, silently drops the new field |

The staleness gap for additive changes is real. The old CI workflow partially caught this (when it worked). Mitigations:
- The `reset-generated-code` script remains available and should be run after API changes.
- A future lightweight CI check could verify just `sdks/` and `docs/` freshness, without regenerating the entire repo.
- Pre-commit hooks can remind developers to regenerate.

### Developer workflow changes

- **Merge conflicts in generated files.** When two branches both trigger generation, developers will see conflicts in files they didn't write. Resolution is always "regenerate and recommit."
- **Generator version skew.** Different local versions of `mockgen` or `swag` can produce different output. Pin versions in `tools.go` to avoid this.
- **Larger PRs.** A small struct change can fan out into dozens of generated files. `linguist-generated` in `.gitattributes` collapses these in GitHub diffs.

### What we chose not to check in

| Pattern | Reason |
|---|---|
| `sdks/*/.swagger.stamp` | Build-system timestamp files, not imported by any code |

## References

| Source | Link |
|---|---|
| Rob Pike, "Generating code" (Go blog) | https://go.dev/blog/generate |
| `go generate` design proposal | https://go.googlesource.com/proposal/+/refs/heads/master/design/go-generate.md |
| Kubernetes contributor guide | https://kubernetes.io/docs/contribute/generate-ref-docs/contribute-upstream/ |
| GitHub Linguist `linguist-generated` docs | https://github.com/github-linguist/linguist/blob/main/docs/overrides.md |
| Jamie Tanna, "Should I commit generated Go code?" | https://www.jvt.me/posts/2022/05/05/commit-go-generate/ |
| templ maintainer recommendation | https://github.com/a-h/templ/discussions/419 |


RealHarshThakur:
Closing, discussed offline on future state of SDKs that can help us elevate this pain.

PR #912, opened 2026-04-13
feat: support sleeping queues + admin commands

jonmorehouse:
This PR automatically puts queue workflows to sleep once they have been idle for a configurable timeout. This reduces 
load from different queues that might be starved of work. 

Since we use `UpdateWithStartWorfklow`, the queue will naturally be restarted when a signal is sent to it, so this is, 
for the most part a free performance optimization for us.


fidiego (approved):
LGTM

PR #913, opened 2026-04-14
fix: auth0 gardening auth regression

erickyellott:
Fixes a fallthrough case introduced when we removed auth0: 

https://github.com/nuonco/nuon/pull/903/changes#diff-6a28a1ae8a0c10936e6b6cba9f6462a284ad74260f0926f3829007020d595b2dL109


fidiego (approved):
LGTM

PR #938, opened 2026-04-15
fix: add context for deadlocker to be paused

RealHarshThakur:
Temporal SDK seems to not pause deadlocking when context passed is nil . 
```
func PauseDeadlockDetector(ctx Context) {
    if d := getDeadlockDetector(ctx); d != nil {  // returns nil for nil ctx
        d.pause()
    }
}
```

The propagator stores the data converter from FX construction, where context is nil. So PauseDeadlockDetector(nil) is a no-op, and json-iterator's first-time reflection cache build for LogStream's deep type graph exceeds the 1-second deadlock detector threshold, causing a temporal workflow panic.


jonmorehouse:
This no longer seems to be an issue with the latest temporal updates.

PR #964, opened 2026-04-17
docs: Add runner auth docs page

erickyellott:
Document how runner authentication works between AWS, GCP, and Azure.


fidiego (approved):
LGTM

PR #972, opened 2026-04-20
feat: install role usgae stats

someshkoli:
This change adds ability to view install role usage for each role and look into it further.

Side bug fixes:
- fixes workflow/v2 signals to accept roles
- original update install stack output to update install roles
- add dev panel for phone home simulation

image


services/dashboard-ui/client/components/admin/dev/PhoneHomeModal/PhoneHomeModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+    try {
+      JSON.parse(displayValue)
+      return true
+    } catch {
+      return false
+    }
+  })()
+
+  return (
+    <ModalBase
```
nnnnat: Unless you need to use ModalBase for some reason this should be Modal

nnnnat (approved):

PR #988, opened 2026-04-21
refactor(api/ui): workflow status description, counts &panel auto-open

nnnnat:
- Updates ctl-api to use step name for workflow status description
- Updates workflow detail page to only poll the workflow and not workflow and steps
- Update workflow detail metrics to show og step counts
- Don't auto-open panels for drift scans


stephennuon (dismissed):

stephennuon (approved):
LGTM

PR #992, opened 2026-04-21
fix: continue as new quicker in process job to avoid accumulating history 

erickyellott:


erickyellott:
gonna try just using the signals instead of optimizing the polling

PR #993, opened 2026-04-21
feat(ctl-api): add step groups to API

jordan-acosta:
## Problem

We added step groups to our data model, but have not exposed them via the api. These would be useful for simplifying client-side code and providing a more polished and consistent UX around workflows.

## Solution

Include step groups in the API.

- Update `GET /workflows/:workflow_id` to include step groups
- Add `GET /v1/workflows/:workflow_id/step-groups`
- Add `GET /v1/workflows/:workflow_id/step-groups/:step_group_id`

### DIFF
--- a/services/ctl-api/internal/app/installs/service/get_workflow.go
+++ b/services/ctl-api/internal/app/installs/service/get_workflow.go
@@ -92,6 +92,12 @@ func (s *service) getWorkflow(ctx *gin.Context, orgID, workflowID string) (*app.
 		Preload("Steps.CreatedBy").
 		Preload("Steps.Approval").
 		Preload("Steps.Approval.Response").
+		Preload("StepGroups", func(db *gorm.DB) *gorm.DB {
+			return db.Order("group_idx asc")
+		}).
+		Preload("StepGroups.Steps", func(db *gorm.DB) *gorm.DB {
+			return db.Order("group_retry_idx, idx, created_at asc")
+		}).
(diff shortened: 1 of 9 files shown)


services/ctl-api/internal/app/installs/service/get_workflow.go
```diff
@@ -92,6 +92,16 @@ func (s *service) getWorkflow(ctx *gin.Context, orgID, workflowID string) (*app.
...
 		Preload("Steps.Approval").
 		Preload("Steps.Approval.Response").
+		Preload("StepGroups", func(db *gorm.DB) *gorm.DB {
+			return db.Order("group_idx asc")
+		}).
+		Preload("StepGroups.Steps", func(db *gorm.DB) *gorm.DB {
+			return db.Order("group_retry_idx, idx, created_at asc")
+		}).
+		Preload("StepGroups.QueueSignal").
+		Preload("StepGroups.Steps.Approval").
```
jonmorehouse: We can't preload steps and approvals here, as they will blow up the queries for these. Those types can be large responses, and are loaded separately intentionally.

jordan-acosta: I'm removing approvals, since I already fetch those from another endpoint. I'm not sure I follow regarding the steps. This endpoint already preloads steps.

Casey-00 (approved):

PR #997, opened 2026-04-22
fix(ctl-api): stable order on AppPolicyConfig preloads

im-Amitto:
## Summary                                                                                                                                                                                                     
  - Add `ORDER BY created_at ASC, id ASC` to all `AppPolicyConfig` preloads so the `Policies` slice is deterministic across runs.
  - Fixes sandbox TF key drift: `plan_sandbox_run.go` maps slice index → `N.yaml`, which Terraform uses as a `for_each` key. Unstable order + `create_before_destroy = true` caused TF to create one object and  
  delete an unrelated one under the same key.                                                                                                                                                                    
                                                                                                                                                                                                                 
  ## Scope                                                                                                                                                                                                       
  Short-term fix. Follow-up commit will replace the positional `N.yaml` key with a content-derived key so it also survives insertions/removals.


services/ctl-api/internal/app/apps/helpers/get_policies_config.go
```diff
@@ -12,7 +13,9 @@ func (h *Helpers) GetPoliciesConfigByAppConfigID(ctx context.Context, appConfigI
 	var policiesConfig app.AppPoliciesConfig
 	res := h.db.WithContext(ctx).
 		Where("app_config_id = ?", appConfigID).
-		Preload("Policies").
+		Preload("Policies", func(db *gorm.DB) *gorm.DB {
+			return db.Order("created_at ASC, id ASC")
```
pcholakov, line 17: Here and elsewhere - is this enough to ensure stable order? Could created_at round to the same value for policies created in the same batch, effectively still falling back on `id`? And even if it was sufficiently distinct, do we guarantee that it's monotonic - and based on what source order?

im-Amitto: All correct: 
1. created_at will tie :- same insert tx, same microsecond. We fall back on id every time in practice.
2. id is a random NanoID, not monotonic, doesn't reflect source order.                                                                                                                                       

What this buys us: stability across re-fetches of the same row set (deterministic tiebreak = same order every query). That's enough to stop TF key drift, since the planner re-reads the same app_config_id and just needs the slice in the same order it saw last apply.

It does not preserve authoring order or survive mid-list inserts. The follow-up commit i mentioned earlier drops the positional N.yaml key for a content-derived one (name/hash) so ordering becomes irrelevant.

PR #1006, opened 2026-04-22
feat: Use signals for ProcessJob instead of polling. 

erickyellott:
Replace Temporal poll loops with signal-based push model to eliminate
`busyworkflow` lock contention.

**Background**: `ProcessJob` polls 4–5 DB state activities every second
while a job runs. Each iteration emits ~3 Temporal history events. A
15-minute job accumulates ~15k events; replaying that MutableState holds
the workflow lock long enough to trigger `busyworkflow` errors in
everything waiting on the same namespace (~4,700/min on stage).


erickyellott:
Closing this for now. Will revisit with Jon's more general changes after stability improvements.

PR #1010, opened 2026-04-24
feat: admin api to retrigger phone home workflow from install stack version run

someshkoli:


services/ctl-api/internal/app/installs/service/admin_trigger_install_stack_output_update.go
```diff
@@ -0,0 +1,43 @@
...
+	var run app.InstallStackVersionRun
+	if res := s.db.WithContext(ctx).
+		Preload("InstallStackVersion").
+		Where("id = ?", runID).
+		First(&run); res.Error != nil {
+		ctx.Error(fmt.Errorf("unable to get install stack version run %s: %w", runID, res.Error))
+		return
+	}
+
+	stackVersion := run.InstallStackVersion
```
someshkoli: port this to new queues system when on main

jonmorehouse (approved):

PR #1088, opened 2026-04-29
chore(ctl-api): use the `latest_runner_heart_beats` table

fidiego:
instead of the `runner_heart_beats` table. this table is populated by
a materized view and de-dedupes on runner_id, process, and process_id.
this means the table effectively stores pre-computed values and is
vastly more efficient.


fidiego:
This is no good. The delay between the time the row makes it into the table via the mv is greater than the time to read it.

PR #1098, opened 2026-04-29
fix: make queue and runner process shutdown more robust

jonmorehouse:
This PR updates the way we handle shut downs to now use a more independent approach. Previously, when we pushed an 
update, we would run a background workflow that updates each event loop one by one. This was expensive, and hard to add 
backoff and jitter into.

This update makes it so we use a "request_continue_as_new" model. This means, we update the object to say it should CAN, 
and then, the queue is responsible for fetching that state and handling it + clearing the update. This should improve 
reliability of this, and at the same time, allow us to more easily scale these types of updates out.

### DIFF
--- a/bins/runner/Dockerfile
+++ b/bins/runner/Dockerfile
@@ -253,8 +253,8 @@ ENV GIT_REF=$tag
 ENTRYPOINT ["/bin/runner"]
 
 FROM scratch AS artifacts-build
-#COPY --from=build-darwin-arm64 /out/ /
-#COPY --from=build-darwin-amd64 /out/ /
+COPY --from=build-darwin-arm64 /out/ /
+COPY --from=build-darwin-amd64 /out/ /
 COPY --from=build-linux-arm /out/ /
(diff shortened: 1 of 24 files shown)


fidiego (dismissed):
LGTM

fidiego (dismissed):

bins/runner/Dockerfile
```diff
@@ -253,8 +253,8 @@ ENV GIT_REF=$tag
 ENTRYPOINT ["/bin/runner"]
 
 FROM scratch AS artifacts-build
-#COPY --from=build-darwin-arm64 /out/ /
-#COPY --from=build-darwin-amd64 /out/ /
```
erickyellott, line 257: what's this?

jonmorehouse: accidentally removed. the darwin builds today. we don't need them, but adding them back in since it was an accident.

fidiego (approved):

PR #1104, opened 2026-04-30
feat: (ctl-api) skip auto-retries on terminal step errors

onprem:
Introduces a generic terminal-error mechanism so signals can mark a step failure as terminal — meaning retrying without external action cannot change the outcome (e.g., missing component build, build in error state). The conductor short-circuits auto-retry and surfaces a stable reason_code for the dashboard / CLI to render.

Without this, a freshly-created install where a component had no successful build would burn the full auto-retry budget (3) on each sync/plan step before failing, even though the underlying condition will never resolve mid-run.

Screenshot 2026-04-30 at 8 16 27 PM

the step in not auto-retried.

Screenshot 2026-04-30 at 8 16 34 PM


services/ctl-api/internal/app/installs/signals/v2/componentdeploysyncandplan/signal.go
```diff
@@ -240,7 +240,14 @@ func (s *Signal) pollForDeployableBuild(ctx workflow.Context, installDeployId, c
 
 		if bld.Status == app.ComponentBuildStatusError {
 			l.Error("component build is in an error state")
-			return fmt.Errorf("component build is in an error state")
+			// Terminal: retrying with the same broken build will never
+			// succeed. Step-level auto-retry short-circuits on this marker
+			// (see signal.IsTerminalError).
+			return signal.NewTerminalError(
+				"component_build_errored",
```
ALPHAvibe: we should probably have a component build terminal error type for these

onprem: So have one defined globally that we can reuse right? Will do.

onprem: done!

RealHarshThakur (approved):

PR #1130, opened 2026-05-01
feat: aws terraform stack support

jordan-acosta:
Re-opening this PR, but with region required for creating AWS installs. We need it to support IID, which is not activated by default in the Cloudformation stack, but is a requirement for some customers.

Since there's no real downside to using IID, the Terraform stack is going to have it activated by default. Vendors can fork the stack to remove it if they want. Customers could also disable it via the tfvars.

## Problem

We want to provide the same Terraform install experience for AWS that we have for GCP. Currently we only support Cloudformation. This has the advantage of providing a (nearly) single-click install experience, but it doesn't work well for vendors or customers who want to manage their infra as code.

## Solution

This PR updates install provision and reprovision workflows to support using the [new Terraform install stack](https://github.com/nuonco/install-stacks/pull/7) for AWS installs.


pkg/runner/auth/aws/iid_certs/us-east-1.pem
```diff
@@ -1,24 +1,18 @@
...
-BgNVHQ4EFgQUrynSPp4uqSECwy+PiO4qyJ8TWSkwgY4GA1UdIwSBhjCBg4AUrynS
-Pp4uqSECwy+PiO4qyJ8TWSmhYKReMFwxCzAJBgNVBAYTAlVTMRkwFwYDVQQIExBX
-YXNoaW5ndG9uIFN0YXRlMRAwDgYDVQQHEwdTZWF0dGxlMSAwHgYDVQQKExdBbWF6
-b24gV2ViIFNlcnZpY2VzIExMQ4IJALFpzEAVWaQZMBIGA1UdEwEB/wQIMAYBAf8C
-AQAwDQYJKoZIhvcNAQELBQADggEBADW/s8lXijwdP6NkEoH1m9XLrvK4YTqkNfR6
-er/uRRgTx2QjFcMNrx+g87gAml11z+D0crAZ5LbEhDMs+JtZYR3ty0HkDk6SJM85
-haoJNAFF7EQ/zCp1EJRIkLLsC7bcDL/Eriv1swt78/BB4RnC9W9kSp/sxd5svJMg
-N9a6FAplpNRsWAnbP8JBlAP93oJzblX2LQXgykTghMkQO7NaY5hg/H5o4dMPclTK
-lYGqlFUCH6A2vdrxmpKDLmTn5//5pujdD2MN0df6sZWtxwZ0osljV4rDjm9Q3VpA
-NWIsDEcp3GUB4proOR+C7PNkY+VGODitBOw09qBGosCBstwyEqY=
```
erickyellott, line 23: Eqy= is the one we want! This is reverting our changes from last night.

erickyellott (changes requested):

pkg/runner/auth/aws/iid_certs/us-east-1.pem
jordan-acosta: oops

erickyellott (approved):

PR #1133, opened 2026-05-02
feat: make plan approvals a signal rather than temporal update

RealHarshThakur:
Wrapped the existing "approve-step" Temporal Update in a new Nuon Signals called workflow-step-approval-request & workflow-step-approval-response so approval wake-ups become first-class queue-backed operations.

With this, we get webhook lifecycle events, queue persistence, and automatic retries for free  instead of a fire-and-forget Temporal Update from the API.


services/ctl-api/internal/app/installs/worker/sandbox/execute_plan.go
```diff
@@ -104,12 +105,18 @@ func (w *Workflows) executeSandboxPlan(ctx workflow.Context, install *app.Instal
 		return errors.Wrap(err, "unable to get job")
 	}
 
-	if _, err := activities.AwaitCreateStepApproval(ctx, &activities.CreateStepApprovalRequest{
-		OwnerID:     sandboxRun.ID,
-		OwnerType:   "install_sandbox_runs",
-		RunnerJobID: job.ID,
-		StepID:      stepID,
-		Type:        app.TerraformPlanApprovalType,
+	var installWorkflowID string
```
someshkoli, line 108: this is old even loop system code, queues have no significance here, these changes are not needed

someshkoli (approved):

PR #1192, opened 2026-05-06
feat: add slack app

RealHarshThakur:
**Install Nuon into a Slack workspace and link it to your orgs.** The dashboard walks you through Slack's OAuth flow, stores the workspace credentials, and lets you link a workspace to one or more orgs (and an org to one or more workspaces).

**Receive notifications wherever you want, scoped to what you care about.** A subscription consists of an event filter (lifecycle, drift, build outcomes, etc.) and a scope. The scope can be the whole org, specific installs, specific components or actions, or a label selector. Both webhooks and Slack channels use this same subscription model, so the scoping vocabulary is identical whether you're piping notifications to a URL or to a channel.

**Pick scope the same way everywhere.** The dashboard editor, the `/nuon subscribe` modal in Slack, and the CLI's interactive picker all present the same flow. Specific picks for components or actions ask for an app first and then narrow within it, since per-entity scoping naturally lives inside an app. Cross-app scoping is expressed with the label selector (for example `env=prod`).

**Register the same destination more than once.** A single webhook URL or Slack channel can hold multiple subscriptions with different scopes and different event filters. For example, one subscription for "everything in prod" and another for "any deploy on the billing app", as separate rows that won't collide.

**Manage subscriptions from anywhere.** New dashboard views list and edit every webhook and Slack channel subscription for an org. The CLI gains named flags for event types and scope on the webhook commands, plus an interactive picker for users who don't want to remember the syntax. From within Slack, `/nuon subscribe` and `/nuon unsubscribe` cover the channel side without leaving the workspace.

**Attach labels to an install at creation time.** The install create call accepts labels in the request body, removing the need for a follow-up call to label a brand-new install. `nuon installs create` accepts `--labels`, and `nuon installs labels` lists, sets, and unsets labels on an existing install.

NOTES:
* Added metrics to capture latency for lifecycle hooks to have better visibility for performance. 
* Short circuited webhook & slack hook paths if no subscribers.


nnnnat (dismissed):

jonmorehouse (dismissed):

stephennuon (dismissed):
LGTM

someshkoli (approved):

someshkoli (approved):

PR #1240, opened 2026-05-08
refactor(api): auth preload & auth validation

nnnnat:
## Summary

Users with many orgs experience slow/failed logins because the ctl-api auth middleware runs expensive database preloads (Roles → Org → Policies) on every API request. The `AfterQuery` hook then iterates all roles × policies to build `OrgIDs` and `AllPermissions`. For accounts with 50+ orgs this is 150+ rows loaded per request.

The BFF login flow makes multiple sequential ctl-api requests, each paying this cost independently.

### ~~1. In-memory TTL cache for account loading~~

~~Added a 30s TTL LRU cache (4096 entries) to `account.Client.FetchAccount` using the existing `hashicorp/golang-lru/v2/expirable` pattern  already used by `RunnerHeartbeatCache`. On cache hit, the 4-table preload query is skipped entirely.~~

~~Accounts with no roles are not cached to avoid stale permissions during the brief window between account creation and role assignment.~~

~~An `InvalidateAccount(id)` method is available for explicit cache busting when roles/permissions are modified.~~

~~**Files:** `services/ctl-api/internal/pkg/account/account.go`, `fetch.go`~~

### 2. Eliminate redundant account load in GetAuthMe

The `GetAuthMe` handler was re-fetching the full account from the database (with all role/policy/org preloads) even though the auth middleware already loaded it into the request context. The only additional data it needed was identity information.

Now reuses the account from context and runs a lightweight query for just the identities table.

**File:** `services/ctl-api/internal/app/accounts/service/get_auth_me.go`

### 3. Lightweight token validation endpoint

Added `GET /v1/auth/validate` — returns `{"valid": true}` if the token is valid, 401 otherwise. No account data serialization, no identity queries.

The BFF auth middleware now calls this instead of `GetAuthMe`, since it only needs to know whether the token is valid and discards the response body. Combined with the account cache, the BFF auth check on every page load typically hits cache instead of running 4+ DB queries.

**Files:**
  - `services/ctl-api/internal/app/accounts/service/validate_token.go` — new endpoint
  - `services/ctl-api/internal/app/accounts/service/service.go` — route registration
  - `services/ctl-api/internal/middlewares/global/global.go` — added to global endpoints list (no org required)
  - `sdks/nuon-go/client.go`, `general.go` — SDK wrapper
  - `services/dashboard-ui/server/internal/middlewares/auth/auth.go` — BFF uses `ValidateToken`

## Trade-offs

- ~~30s cache TTL means permission changes take up to 30s to take effect. Acceptable since permission changes are rare and the client-side polling interval is already 30s.~~
- ~~Accounts with no roles bypass the cache entirely to prevent stale empty-permission entries during account creation + role assignment.~~

## Test plan

- [x] Existing Go tests pass (`go test ./services/ctl-api/...`)
- [x] Manual: login with multi-org account, confirm faster load
- [x] Manual: fresh account signup flows through to onboarding
- [x] Manual: `GET /v1/auth/validate` returns 200 with valid token, 401 without
- [x] Verify permission changes propagate within ~30s


fidiego (approved):
LGTM

PR #1252, opened 2026-05-11
fix: hard error on unknown signal

someshkoli:
when a new signal is introduced in db and then is removed from code, multiple code paths fail because Scan method hard fails on signal not found.
Instead of hard failing, we now log the failure and continue forward and rely on signal nil check for further operations.
:x

### DIFF
--- a/services/ctl-api/internal/pkg/queue/catalog/obj.go
+++ b/services/ctl-api/internal/pkg/queue/catalog/obj.go
@@ -4,14 +4,14 @@ import (
 	"errors"
 	"fmt"
 
+	"go.temporal.io/sdk/temporal"
+
 	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/queue/signal"
 )
 
-// ErrSignalTypeNotRegistered is returned by NewFromType when the given signal
-// type is not present in the catalog. Callers can use errors.Is to detect this
-// specific condition — e.g. to gracefully skip stale DB rows that reference
-// deprecated signal types instead of failing the whole query.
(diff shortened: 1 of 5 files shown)


services/ctl-api/internal/pkg/queue/catalog/obj.go
```diff
@@ -1,15 +1,22 @@
...
+// type is not present in the catalog. Callers can use errors.Is to detect this
+// specific condition — e.g. to gracefully skip stale DB rows that reference
+// deprecated signal types instead of failing the whole query.
+var ErrSignalTypeNotRegistered = errors.New("signal type not registered")
+
 func NewFromType(typ signal.SignalType) (signal.Signal, error) {
 	constructor, ok := SignalCatalog[typ]
 	if !ok {
-		return nil, fmt.Errorf("invalid signal type %s - not registered", typ)
+		return nil, fmt.Errorf("%w: %s", ErrSignalTypeNotRegistered, typ)
```
ALPHAvibe, line 19: the workers doesn't end up in a retry loop right?

someshkoli: so that case will only be possible if a workflow is active post rollback, which is possible, leme patch that

someshkoli: addressed, on activity we are returning non retryable error, and in workflow code we are requeueing the signal, @jonmorehouse i'm guessing this is mostly fine since even if requeued it'll get marked as pending again until it has a queue signal in the catalog to work with.

ALPHAvibe (approved):

PR #1254, opened 2026-05-11
feat: Composite Errors: Add GORM models and activity/helpers to read and write errors from steps

onprem:
This PR adds 
- The GORM models: with indexes, constraints etc
- Helpers and Activities: to read error trees, and build context from steps for error parsing, and then to writing them in DB

This is everything that we need to wire up Composite Errors in the error processing logic for Workflow Steps.


RealHarshThakur (approved):

services/ctl-api/internal/app/composite_errors/activities/activities.go
```diff
@@ -0,0 +1,29 @@
...
+// CompositeError without depending on the database directly.
+//
+// The Helpers package is the single source of write logic; activities are a
+// thin RPC boundary around it.
+package activities
+
+import (
+	"go.uber.org/fx"
+
+	cehelpers "github.com/nuonco/nuon/services/ctl-api/internal/app/composite_errors/helpers"
```
jonmorehouse, line 12: This package should be in `internal/pkg/compositeerrors` similar to where we put things like `blob` and others.

onprem:
closed in favour of #1274

PR #1265, opened 2026-05-11
chore(dashboard-ui): pin tan stack deps

fidiego:
- **chore: config changes'**
- **chore: pin node modules**


nnnnat:
These are all pinned now

PR #1292, opened 2026-05-12
fix: optimize install query preloads  

erickyellott:
* Add limit to InstallSandboxRuns Preload
* Remove broken batching


services/ctl-api/internal/app/installs/service/get_org_installs.go
```diff
@@ -100,55 +108,10 @@ func (s *service) getOrgInstalls(ctx *gin.Context, orgID, q string, lbls labels.
...
-			Preload("InstallDeploys", func(db *gorm.DB) *gorm.DB {
-				return db.
-					Scopes(scopes.WithOverrideTable(views.CustomViewName(s.db, &app.InstallDeploy{}, "latest_view_v1"))).
-					Order("install_deploys_latest_view_v1.created_at DESC")
-			}).
-			Preload("Component").
-			Where("install_id = ?", install.ID).
-			Limit(batchSize).
-			Offset(offset).
-			Find(&installComponents)
```
erickyellott, line 141: @ALPHAvibe `Find(&installComponents)` should have been `Find(&installComponentsBatch)`, this was only ever loading the final batch. 

> // WARN: (rb) Get install components in batches to avoid loading too many components into memory at once

But if we're just fetching and storing things into memory, I don't see how batching would have fixed any memory issues here.

PR #1332, opened 2026-05-14
feat: readme details endpoints

ALPHAvibe:


services/ctl-api/internal/app/runners/service/admin_list_runners_details.go
```diff
@@ -128,6 +151,12 @@ func (s *service) listRunnersDetails(ctx *gin.Context, limit, offset int) ([]*Ad
 	for _, r := range runners {
 		item := &AdminRunnerDetails{
 			Runner:            r,
+			Type:              r.RunnerGroup.Type,
```
ALPHAvibe, line 154: these might be redundent since we always preload

jordan-acosta (approved):

PR #1339, opened 2026-05-15
feat: create org cleanup dashboard

jonmorehouse:
Creating a dashboard that allows for cleaning up an org:

1. forget installs
1. remove org infra
1. remove queues
1. remove queue signals
1. remove temporal workflows

### DIFF
--- a/services/ctl-api/internal/app/admin-dashboard/service/org_queue_signals.go
+++ b/services/ctl-api/internal/app/admin-dashboard/service/org_queue_signals.go
@@ -1,0 +1,48 @@
+package service
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"go.uber.org/zap"
+
+	"github.com/nuonco/nuon/services/ctl-api/internal/app"
+)
+
+// OrgQueueSignals returns all queue signals for an org.
+func (s *service) OrgQueueSignals(c *gin.Context) {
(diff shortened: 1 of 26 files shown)


services/ctl-api/internal/app/admin-dashboard/service/org_queue_signals.go
```diff
@@ -0,0 +1,48 @@
...
+		Find(&signals); res.Error != nil {
+		s.l.Error("failed to list org queue signals", zap.Error(res.Error), zap.String("org_id", orgID))
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to list queue signals"})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{"signals": signals})
+}
+
+// DeleteOrgQueueSignals hard-deletes all queue signals for an org.
```
ALPHAvibe, line 30: does the delete have a confirmation dialogue?

ALPHAvibe (dismissed):

services/ctl-api/internal/app/admin-dashboard/service/org_queue_signals.go
jonmorehouse: no, but I'll follow up on that. good call out

ALPHAvibe (approved):

PR #1347, opened 2026-05-15
refactor(ui-): step status on logs, outputs on workflows, fix dev-script

nnnnat:
- Step status & log count in action logs
- Show action run outputs in workflow panel
- fix issue with action run log filters
- fix issue with dev script
- better missing diff message


fidiego (approved):
LGTM

PR #1353, opened 2026-05-17
feat(cli): nuon-ext-replay-stack cli extension

jordan-acosta:
## Problem

If a stack phone home is not received while it's reprovision workflow is running, then the stack status will be left in an "awaiting-user-run" state, even if the actual stack if fully-provisioned and the runner may be healthy. This blocks the vendor from running most operations.

Currently, the only solution is to have the customer re-run. the stack to re-post the phone home request, which is a bad look for the vendor and annoying to customers.

## Solution

Since the API just needs a POST request with the right outputs, we can empower vendors to do the following:

1. Run a new reprovision workflow and wait for it to get to the "await stack run" step.
1. Read the outputs from the most recent stack run that succeeded.
2. POST those outputs to the new phone home URL.

The status will be updated, the reprovision workflow can complete, and the vendor will be unblocked without having to hassle their customer.

The easiest way to allow the vendor to do this is with a CLI extension. The actual extension is implemented in https://github.com/nuonco/nuon-ext-replay-stack. This PR fixes a couple issues I found while iterating on the extension:

- Fixes a bug that's excluding locall-installed extensions from `nuon ext list`
- Updates a local README to correctly document that you no longer need `NUON_PREVIEW` to use `nuon ext`. LLMs will be confused by this and tell you it's still required.


fidiego (approved):
LGTM

PR #1370, opened 2026-05-18
feat: make mng shutdown job work directly

jonmorehouse:
This PR makes the legacy shutdown vm command work by creating a ready-to-process job.


fidiego (approved):
LGTM

PR #1381, opened 2026-05-19
fix(queue): bound retry on best-effort signal status updates

im-Amitto:
Without explicit bounds, the in-progress / post-error UpdateQueueSignalStatusV2 calls retry indefinitely when the signal row is missing (e.g. stale-dropped by emitter), holding the queue's MaxInFlight slot forever. Cap to 3 attempts / 2 min.


im-Amitto:
Superseded by bumping MaxInFlightAge instead — addresses the root cause (stale threshold too aggressive) rather than patching the symptom (unbounded retry).

PR #1404, opened 2026-05-20
chore: support parallel enqueuing via channel

jonmorehouse:
This PR supports parallel, and configurable enqueing of signals.

Previously, the enqueuer channel would only handle a single signal at a time, this updates to use a worker model and handle up to a configurable number of signals in flight.


stephennuon (approved):
Promotion time 🚀

PR #1410, opened 2026-05-20
fix: config refs docs

someshkoli:
fix config refs docs generation to deduplicate pulumi config across rest
of the configurations


matt-zach-s (approved):
LGTM

PR #1414, opened 2026-05-21
feat(dashboard): app branch design updates

stephennuon:
### Summary

- Switch the private repo badge to theme="default" so it's actually legible inside the Repository dropdown (was falling back to low-contrast neutral).
- Drop the preemptive Git-branch skeleton that rendered while only repositories were loading — branches can't load before a repo is picked, so the second skeleton was misleading.
- Keep real Repository and Git branch labels visible during loading and only skeletonize the input field, eliminating label flicker when data lands.


fidiego (approved):
LGTM

PR #1501, opened 2026-05-28
feat: workflow search

jordan-acosta:
## Problem

We can't search on the install workflow history page like we can on other install pages.

## Solution

This PR adds a search box and updates the layout a bit to bring it in line with other install pages with searchable lists. Because workflows are paginated server-side, and the workflow names are created dynamically client-side, search had to be implemented in a different way from other pages.

### Adding server-side search by name

A new `search` new parameter was added to the workflows GET endpoint. The query was updated with a SQL expression duplicating the client-side naming logic, to support searching for the names you see in the UI.

Screenshot 2026-05-27 at 23 02 28

### Moving workflow naming server-side

To dedupe the naming logic, the SQL expression was moved to a stored generated column, and the UI updated to use this instead of composing the name itself.

```sql
ALTER TABLE install_workflows
ADD COLUMN name TEXT GENERATED ALWAYS AS (
	CASE
	WHEN type = 'action_workflow_run' AND COALESCE(metadata->'adhoc_action', '') <> ''
		THEN 'Adhoc action run (' || COALESCE(metadata->'install_action_workflow_name', '') || ')'
	ELSE
		COALESCE(
			CASE
			WHEN finished_at IS NULL THEN
				CASE type
					WHEN 'provision' THEN 'Provisioning install'
					WHEN 'reprovision' THEN 'Reprovisioning install'
					WHEN 'deprovision' THEN 'Deprovisioning install'
					WHEN 'manual_deploy' THEN 'Deploying to install'
					WHEN 'drift_run' THEN 'Deploying to install'
					WHEN 'input_update' THEN 'Input Update'
					WHEN 'teardown_components' THEN 'Tearing down all components'
					WHEN 'deploy_components' THEN 'Deploying all components'
					WHEN 'reprovision_sandbox' THEN 'Reprovisioning sandbox'
					WHEN 'drift_run_reprovision_sandbox' THEN 'Reprovisioning sandbox'
					WHEN 'sync_secrets' THEN 'Syncing secrets'
					WHEN 'action_workflow_run' THEN 'Action run'
					WHEN 'app_config_build' THEN 'Building app config components'
					WHEN 'runbook_run' THEN 'Running runbook'
				END
			ELSE
				CASE type
					WHEN 'provision' THEN 'Provisioned install'
					WHEN 'reprovision' THEN 'Reprovisioned install'
					WHEN 'reprovision_sandbox' THEN 'Reprovisioned sandbox'
					WHEN 'drift_run_reprovision_sandbox' THEN 'Reprovisioned sandbox'
					WHEN 'deprovision' THEN 'Deprovisioned install'
					WHEN 'manual_deploy' THEN 'Deployed to install'
					WHEN 'drift_run' THEN 'Deployed to install'
					WHEN 'input_update' THEN 'Updated Input'
					WHEN 'teardown_components' THEN 'Tore down all components'
					WHEN 'deploy_components' THEN 'Deployed all components'
					WHEN 'sync_secrets' THEN 'Synced secrets'
					WHEN 'action_workflow_run' THEN 'Action run'
					WHEN 'app_config_build' THEN 'Built app config components'
					WHEN 'runbook_run' THEN 'Runbook run'
				END
			END,
			REPLACE(type, '_', ' ')
		)
		|| COALESCE(' (' || NULLIF(metadata->'workflow-name-suffix', '') || ')', '')
		|| CASE WHEN type = 'action_workflow_run'
				THEN COALESCE(' (' || NULLIF(metadata->'install_action_workflow_name', '') || ')', '')
				ELSE '' END
		|| CASE WHEN type = 'runbook_run'
				THEN COALESCE(' (' || NULLIF(metadata->'runbook_name', '') || ')', '')
				ELSE '' END
	END
) STORED;
```


services/ctl-api/internal/app/workflow.go
```diff
@@ -257,7 +261,13 @@ func (i *Workflow) BeforeCreate(tx *gorm.DB) error {
 	return nil
 }
 
+// BeforeSave keeps install_workflows.name in sync with computeWorkflowName.
+// Callers that mutate type, metadata, or finished_at must pass a fully
+// populated struct (load-then-Save) so this hook can see the post-update
+// state. Partial Updates(map / struct-with-only-finished_at) and raw SQL
+// bypass the hook's view and will leave name stale.
 func (i *Workflow) BeforeSave(tx *gorm.DB) error {
+	i.Name = computeWorkflowName(i)
```
ALPHAvibe, line 270: should we have this in beforecreate as well?

jordan-acosta: ~~yes we should, I thought I added that~~ Actually, wait, I started with `BeforeCreate`, but `BeforeSave` is called before both creating and saving, so I switched.

https://gorm.io/docs/hooks.html#Creating-an-object

nnnnat (approved):

services/ctl-api/internal/app/workflow.go
ALPHAvibe: ahh your right, mixed it up with beforeupdate

ALPHAvibe (approved):

PR #1502, opened 2026-05-28
fix: remove created by inheritance for component builds via component

someshkoli:
This bug causes all component build to go under the name of compoennt
creator. Causing all component builds to have same creator name post
component was created.


RealHarshThakur (dismissed):

someshkoli:
stale pr

PR #1513, opened 2026-05-28
chore: stop sending legacy Slack webhook notifications

RealHarshThakur:
The platform no longer posts to the legacy internal Slack webhook. Every signal that previously fanned out to both email and the webhook now only sends email. The Slack app integration (channel subscriptions, slash command, threaded posts) is unchanged and remains the supported way to get Slack notifications.


ALPHAvibe (approved):

jonmorehouse (approved):

jonmorehouse:
Thanks, @RealHarshThakur - are we also able to remove the legacy email notifications package as well? We can move transactional emails into the event stream.

PR #1541, opened 2026-06-01
feat: (ctl-api) enforce 5-minute minimum interval for action and drift crons

onprem:
Adds a `MinCronInterval` (5m) check to the shared `cron_schedule` validator so action triggers, component drift schedules, and sandbox drift schedules are all rejected when they would fire more often than once every 5 minutes.

- Extend `cronScheduleValidator` to walk the next 200 fires and reject when the smallest interval is below `MinCronInterval`. Using this approach as crons schedules can fire on uneven intervals.
- Tag `DriftSchedule` on terraform/helm/pulumi/k8s component configs and on the app sandbox config with 'omitempty,cron_schedule' and drop the inline cron.ParseStandard blocks now covered by struct validation.
- Mirror the rule in syncer/validation.ValidateCronSchedule, and call it from the action and sandbox syncer paths which write directly to the DB without going through the HTTP validators.
- Add unit tests for the validator covering pass cases (>= 5m), too-frequent schedules (incl. mixed-interval crons with one short gap), and bad syntax.

### DIFF
--- a/services/ctl-api/internal/app/components/service/create_helm_component_config.go
+++ b/services/ctl-api/internal/app/components/service/create_helm_component_config.go
@@ -11,7 +11,6 @@ import (
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/lib/pq"
 	"github.com/pkg/errors"
-	"github.com/robfig/cron"
 
 	"github.com/nuonco/nuon/services/ctl-api/internal/app"
 	"github.com/nuonco/nuon/services/ctl-api/internal/middlewares/stderr"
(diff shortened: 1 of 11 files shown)


services/ctl-api/internal/app/components/service/create_helm_component_config.go
```diff
@@ -234,10 +233,6 @@ func (s *service) createHelmComponentConfig(ctx context.Context, cmpID string, r
 		OperationRoles:               operationRoles,
 	}
 	if req.DriftSchedule != nil {
-		_, err := cron.ParseStandard(*req.DriftSchedule)
```
ALPHAvibe, line 237: why is this removed?

ALPHAvibe (approved):

PR #1559, opened 2026-06-02
fix(installs): apply input defaults when an install has no values set

matt-zach-s:
## Summary

`ToInputState` (`services/ctl-api/internal/app/installs/helpers/get_install_state.go`) returned a **nil** `*state.InputsState` whenever an install's inputs record had **zero values set**, so the app config's input **defaults were never materialized**. The render context's `.nuon.inputs` was then `nil`, and any component / README / runbook template that references `.nuon.inputs.inputs.*` panicked during planning:

```
nil pointer evaluating interface {}.inputs
```

This is independent of the app — any app whose components interpolate an input value crashes to deploy if an install is created without explicitly setting at least one input. It surfaced on a `whoami`-on-EKS config whose ALB helm component sets `domain = "{{ .nuon.inputs.inputs.subdomain }}.{{ … }}"`.

### Root cause

```go
func ToInputState(inputs *app.InstallInputs, cfg *app.AppConfig, redacted bool) *state.InputsState {
	if inputs == nil { return nil }
	inputValues := inputs.Values
	if redacted { inputValues = inputs.ValuesRedacted }
	if len(inputValues) < 1 {
		return nil           // ← BUG: skips default materialization entirely
	}
	is := state.NewInputsState()
	for _, inp := range cfg.InputConfig.AppInputs {
		val, ok := inputValues[inp.Name]
		if !ok { val = &inp.Default }   // defaults only applied *past* the early-return
	}
	...
}
```

Defaults are only applied inside the loop, which is unreachable when no values are set. So a brand-new install (no inputs provided) gets `.nuon.inputs == nil` instead of `{subdomain: "whoami", …}`.

---

### A. Self-contained reproduction (no infra, compiles on `main`)

`pkg/render` is unaffected by the bug, so this models exactly what the renderer receives for an install whose inputs record has no values (`ToInputState` → `nil` → `.nuon.inputs == nil`). On `main` it panics; with this PR `ToInputState` returns the defaults so `.nuon.inputs` is populated.

Save as `cmd/repro/main.go` on `main` and `go run ./cmd/repro`:

```go
package main

import (
	"fmt"

	"github.com/nuonco/nuon/pkg/render"
)

func main() {
	// Install render context when ToInputState returns nil (no input values set):
	st := map[string]any{
		"nuon": map[string]any{
			"inputs": nil, // <-- ToInputState(...) == nil on main
			"install": map[string]any{
				"sandbox": map[string]any{"outputs": map[string]any{
					"nuon_dns": map[string]any{"public_domain": map[string]any{"name": "ex.nuon.run"}},
				}},
			},
		},
	}
	// A helm component value, exactly like example-app-configs/eks-simple-auto's alb.toml:
	values := map[string]any{
		"domain": "{{ .nuon.inputs.inputs.subdomain }}.{{ .nuon.install.sandbox.outputs.nuon_dns.public_domain.name }}",
	}
	if err := render.RenderMap(&values, st); err != nil {
		fmt.Println("RENDER FAILED:", err)
		return
	}
	fmt.Println("rendered:", values["domain"])
}
```

Output on `main`:

```
RENDER FAILED: unable to render interface string map value: unable to execute template:
template: input:1:8: executing "input" at <.nuon.inputs.inputs.subdomain>:
nil pointer evaluating interface {}.inputs
```

### B. Unit-level demonstration

The added test `TestToInputState_AppliesDefaultsWhenNoValuesSet` fails on `main` (it asserts a non-nil `InputsState` with defaults applied; `main` returns `nil`) and passes with this PR:

```bash
go test ./services/ctl-api/internal/app/installs/helpers/ -run TestToInputState -v
```

### C. End-to-end (what was actually hit)

1. Sync any app whose component interpolates an input, e.g. `eks-simple-auto` (its `alb.toml` uses `{{ .nuon.inputs.inputs.subdomain }}`).
2. Create an install **without setting any inputs** (accept the defaults; the dashboard/CLI does not force you to set them).
3. Provision, then deploy the `application_load_balancer` component.
4. The plan workflow fails: `unable to render map: … <.nuon.inputs.inputs.subdomain>: nil pointer evaluating interface {}.inputs` (`createHelmDeployPlan` → `render.RenderStruct` → `render.RenderMap`).

---

## Fix

Build the `InputsState` from the app config's declared inputs **regardless of how many values are set**, overlaying any set values on top of the defaults — i.e. drop the `len(inputValues) < 1` early-return. Installs that set ≥1 value are unaffected (same code path as before).

## Tests

- `TestToInputState_AppliesDefaultsWhenNoValuesSet` — reproduces the bug (record present, zero values) → expects defaults applied.
- `TestToInputState_OverlaysSetValuesOnDefaults` — guards the existing behavior (set values override defaults; unset inputs fall back to defaults).

## Verification notes

- The fix and tests are `gofmt`-clean.
- The render-level reproduction (A) was run before/after and confirms the panic disappears (`whoami.ex.nuon.run`) once `.nuon.inputs` is non-nil.
- ⚠️ Heads-up for reviewers running tests locally: on a fresh `main` checkout the `helpers` package currently **fails to compile** due to **pre-existing stale `temporal-gen` output** in `internal/pkg/queue/handler/activities` (`update_workflow__*_gen.go` out of sync with their hand-written activities — the "run `go generate`" situation noted in `AGENTS.md`), unrelated to this change. Regenerate temporal code (or rely on CI's generated tree) before `go test` on this package.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matt-zach-s:
closing - known bug

PR #1577, opened 2026-06-03
fix: context propagation into signal handler

someshkoli:
Queue workflow does not propagate signal context to handler workflow which causes improper information in handler workflow context.
It also propagates queue workflow context further which could cause incorrect ctx values inside handler values.

This pr applies signal context before creating handler workflow so correct ctx values are propagated further.

### DIFF
--- a/services/ctl-api/internal/app/components/worker/activities/create_component_build_record.go
+++ b/services/ctl-api/internal/app/components/worker/activities/create_component_build_record.go
@@ -11,7 +11,6 @@ import (
 type CreateComponentBuildRecordRequest struct {
 	ComponentID string `validate:"required"`
 	OrgID       string `validate:"required"`
-	CreatedByID string
 
 	// GitRef overrides useLatest when set. Used to pin a build to the branch's specific commit.
 	GitRef *string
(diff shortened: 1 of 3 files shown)


onprem (dismissed):

services/ctl-api/internal/app/components/worker/activities/create_component_build_record.go
```diff
@@ -25,7 +25,9 @@ type CreateComponentBuildRecordRequest struct {
 // @temporal-gen-v2 activity
 func (a *Activities) CreateComponentBuildRecord(ctx context.Context, req CreateComponentBuildRecordRequest) (*app.ComponentBuild, error) {
 	ctx = cctx.SetOrgIDContext(ctx, req.OrgID)
-	ctx = cctx.SetAccountIDContext(ctx, req.CreatedByID)
+	if req.CreatedByID != "" {
```
ALPHAvibe: why is this conditional?

someshkoli: I just left that in, in-case caller sends specific account, i don't think there's a usage of it tho, didn't audit entire thing.

someshkoli: removed

ALPHAvibe (approved):

PR #1579, opened 2026-06-03
refactor(api): update how we handle runbook delete and config loading

nnnnat:
- Don't show install runbooks if the runbook was deleted from the app
- Preload the config/runs info for runbooks so the latest run and step count is correct every where


ALPHAvibe (approved):

stephennuon (approved):
LGTM

PR #1588, opened 2026-06-03
chore: properly wire timeouts between steps, groups and signals

jonmorehouse:
We were inadvertently using a fallback timeout of 30 minutes in some places. This updates the timeouts to propagate 
correctly:

1. step-groups correctly calculate the aggregate timeout they need to set.
1. steps propagate the timeout from inner signals.
1. jobs correctly set timeouts.
1. no more `Await` method, and we enforce callers pass in a timeout.

There are some places we were setting smaller timeouts than needed, or larger timeouts than needed.

### DIFF
--- a/services/ctl-api/internal/app/installs/helpers/stategen/stategen.go
+++ b/services/ctl-api/internal/app/installs/helpers/stategen/stategen.go
@@ -1,8 +1,6 @@
 package stategen
 
 import (
-	"time"
-
 	"github.com/pkg/errors"
 	"go.temporal.io/sdk/workflow"
 
@@ -53,7 +51,7 @@ func HintOrGenerate(ctx workflow.Context, req Request) error {
 			}
 			// state-manager queue missing — fall through to legacy generation
 		} else {
-			if _, err := callback.AwaitWithTimeout(ctx, cb, 30*time.Minute); err != nil {
+			if _, err := callback.AwaitWithTimeout(ctx, cb, callback.ShortTimeout); err != nil {
 				return errors.Wrap(err, "unable to await state generation")
 			}
 			return nil
--- a/services/ctl-api/internal/app/installs/helpers/stategen/stategen.go
+++ b/services/ctl-api/internal/app/installs/helpers/stategen/stategen.go
@@ -1,6 +1,8 @@
 package stategen
 
 import (
+	"time"
+
 	"github.com/pkg/errors"
 	"go.temporal.io/sdk/workflow"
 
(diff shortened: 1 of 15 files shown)


services/ctl-api/internal/app/installs/signals/driftdetected/dispatch.go
```diff
@@ -31,7 +33,7 @@ func Dispatch(ctx workflow.Context, sig *Signal) error {
 		return errors.Wrap(err, "unable to enqueue drift-detected signal")
 	}
 
-	if _, err := callback.Await(ctx, cb); err != nil {
+	if _, err := callback.AwaitWithTimeout(ctx, cb, 5*time.Minute); err != nil {
```
someshkoli: lets keep them in const variable as we are using them in multiple places

services/ctl-api/internal/app/installs/signals/workflowstepapprovalrequest/signal.go
```diff
@@ -152,7 +154,7 @@ func Dispatch(ctx workflow.Context, sig *Signal) error {
 		return errors.Wrap(err, "unable to enqueue workflow-step-approval-request signal")
 	}
 
-	if _, err := callback.Await(ctx, cb); err != nil {
+	if _, err := callback.AwaitWithTimeout(ctx, cb, 5*time.Minute); err != nil {
```
someshkoli: same here

services/ctl-api/internal/app/installs/helpers/stategen/stategen.go
```diff
@@ -51,7 +53,7 @@ func HintOrGenerate(ctx workflow.Context, req Request) error {
 			}
 			// state-manager queue missing — fall through to legacy generation
 		} else {
-			if _, err := callback.Await(ctx, cb); err != nil {
+			if _, err := callback.AwaitWithTimeout(ctx, cb, 30*time.Minute); err != nil {
```
someshkoli: same here

services/ctl-api/internal/pkg/queue/handle_signal.go
```diff
@@ -121,11 +122,13 @@ func (q *queue) processQueueSignal(ctx workflow.Context, l *zap.Logger, queueSig
 		return errors.Wrap(err, "unable to send validate update")
 	}
 
-	if _, err := callback.Await(ctx, validateCB); err != nil {
+	if _, err := callback.AwaitWithTimeout(ctx, validateCB, 5*time.Minute); err != nil {
```
someshkoli: same

PR #1589, opened 2026-06-04
fix: cli handles  api urls with end slashes

jordan-acosta:
## Problem

Some of the new SDK endpoints can't handle a configured `api_url` with a trailing slash.

## Solution

If there is a slash at the end of the API URL, strip it before appending the path.


erickyellott (approved):

fidiego (approved):
LGTM

PR #1595, opened 2026-06-04
fix: update entity related endpoints to use latest app configs

someshkoli:


services/ctl-api/internal/app/runbooks/service/get_runbooks.go
```diff
@@ -45,6 +45,22 @@ func (s *service) GetRunbooks(ctx *gin.Context) {
 	q := ctx.Query("q")
 	lbls := labels.ParseLabelsQuery(ctx.Query("labels"))
 
+	var currentAppConfig app.AppConfig
```
someshkoli, line 53: known repetition, will take it out if we have one more instance of this

PR #1606, opened 2026-06-05
fix: batch sync ensure actions

ALPHAvibe:
we have unbounded handling of installs and their actions.
this change batches ensuring we only handle 20 records at a time.


someshkoli (approved):

onprem (approved):

fidiego (approved):
LGTM

PR #1629, opened 2026-06-08
feat: welcome a channel when the Slack bot joins

RealHarshThakur:
When the Nuon bot is added to a Slack channel, it now posts a one-time welcome that greets people with the bot's own display name and points them to subscribe and help.
Slash command help now echoes the actual command name the workspace uses, instead of a fixed one.


fidiego (approved):
LGTM

PR #1639, opened 2026-06-09
feat: Add CompositeError to install component deploy jobs

onprem:
This PR adds composite error support to deploy jobs for install components. 

Currently only terraform permission related errors have been wired up to be extracted out and shown to users. We detect AWS permission errors in case of Component Plan and Deploy jobs. I'll follow up with wiring this to also detect AWS permission errors in case of Sandbox Plan and Apply steps as well as for Action runs.

In the UI, we have added a new component for `CompositeError` and we render it on deploy job's detail page, as well as on step details page.

Screenshot 2026-06-09 at 9 31 05 PM

### DIFF
--- a/services/ctl-api/internal/app/installs/deployerrors/aws_permission_error.go
+++ b/services/ctl-api/internal/app/installs/deployerrors/aws_permission_error.go
@@ -1,0 +1,118 @@
+// Package deployerrors holds the typed CompositeError implementations produced
+// by component deploy (plan + apply) flows. It lives next to the installs
+// domain that consumes it rather than in a central error catalog: each domain
+// owns the custom errors it knows how to produce and parse.
+package deployerrors
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/compositeerrors"
+)
+
+// AWSPermissionErrorType is the discriminator for AWS IAM permission failures
+// surfaced from a terraform plan/apply during a component deploy.
+const AWSPermissionErrorType compositeerrors.Type = "terraform.aws_permission"
+
+// AWSPermissionError is the typed payload for an AWS API call that failed with
+// AccessDenied / UnauthorizedOperation because the deploy's IAM principal is
+// missing a permission. It implements compositeerrors.CompositeError so it can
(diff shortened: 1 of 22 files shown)


services/ctl-api/internal/app/installs/deployerrors/aws_permission_error.go
```diff
@@ -0,0 +1,118 @@
...
+		{
+			Heading: "Why",
+			Body:    "The IAM principal used by this deployment is missing a permission required to perform the operation. Grant the permission to the principal and retry.",
+		},
+	}
+
+	if e.RawMessage != "" {
+		sections = append(sections, compositeerrors.Section{
+			Heading: "AWS response",
+			Body:    "```\n" + e.RawMessage + "\n```",
```
ALPHAvibe, line 75: why the leading \n

services/ctl-api/internal/app/installs/deployerrors/aws_permission_error.go
```diff
@@ -0,0 +1,118 @@
...
+
+// policyStatementJSON renders a minimal IAM policy statement granting the
+// missing action on the resource (or "*" if the resource isn't known).
+func (e *AWSPermissionError) policyStatementJSON() string {
+	resource := e.Resource
+	if resource == "" {
+		resource = "*"
+	}
+	stmt := map[string]any{
+		"Version": "2012-10-17",
```
ALPHAvibe: can we ref this https://github.com/nuonco/nuon/blob/a5398020d9168f6345a4a773e81eb9edcb4d2c19/services/ctl-api/internal/app/orgs/worker/iam/roles/deployments.go#L8

to avoid having to update it if there are changes.

services/ctl-api/internal/app/installs/deployerrors/aws_permission_error.go
onprem: these are to keep the code blocks tag (```) in it's own line so the block renders correctly.

onprem: Those `\n` are sandwiched between the markdown code tag.

ALPHAvibe (approved):

PR #1642, opened 2026-06-09
refactor(ui): edit labels in install quick menu

nnnnat:
Adds "Edit labels" to the install quick menu. Refresh installs query when labels are updated.


stephennuon (approved):
send it

PR #1652, opened 2026-06-10
feat(ui): byoc stepup flow/page

nnnnat:
Add a /byoc-setup route to be used instead of the general onboarding flow for byoc installs of nuon


stephennuon (approved):
Let's go to Market.

PR #1655, opened 2026-06-10
fix: azure runner use binary health probes and don't shutdown immediately

erickyellott:
A shutdown on the runner in Azure will replace the instance, but it takes 10+ minutes to do. 

Here we:
* Swap the health probes to binary mode (dumb 200 OK healthchecks)
* Don't call shutdown ourselves on Azure runners
* Instead mark the azure healthcheck as unhealthy and leave the runner up
* Allow the Azure control plane to replace the VM (in its own sweet time)


erickyellott:
On further testing, on shutdown, the VMSS keeps the last reported health state and doesn't actually roll the VM: 
image

erickyellott:
image
OK this does roll, but it takes a long time. Not sure what the time frame is, but it seems longer than the 10m grace period. Maybe ~20m to finally come back up.

jordan-acosta (dismissed):

jordan-acosta (approved):

PR #1657, opened 2026-06-11
chore: stop useless code comments

nnnnat:
Update agents.md to be more clear about frivolous code comments from agents.


someshkoli (approved):
💯

PR #1659, opened 2026-06-11
refactor(ui): minus icon for empty table columns

nnnnat:
Make sure empty table columns show the minus element


stephennuon (approved):
LGTM. Send it.

PR #1664, opened 2026-06-11
refactor(api): optimize workflow step preloads

nnnnat:
- get_workflow.go — dropped the StepGroups.Steps.* preloads (steps no longer loaded/serialized twice; step_groups is metadata only) and added Omit("contents") to Steps.Approval
- get_workflows.go, get_org_workflows.go — Omit("contents") on Steps.Approval
- get_workflow_step_groups.go, get_workflow_step_group.go — Omit("contents") on Steps.Approval
- get_org_pending_approvals.go — Omit("contents") on the pending-approvals view query
- get_workflow_step_approval.go — left loading contents (the contents-download endpoint reads through it); comment added
- New tests: TestGetWorkflowOmitsGroupStepsAndApprovalContents, TestGetOrgPendingApprovalsOmitsContents

Contract impact: step_groups[].steps disappears from GET /v1/workflows/{id} (no consumers found); everything else is byte-identical since contents was never serialized.

### DIFF
--- a/services/ctl-api/internal/app/installs/service/get_org_workflows.go
+++ b/services/ctl-api/internal/app/installs/service/get_org_workflows.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/pkg/errors"
-	"gorm.io/gorm"
 
 	"github.com/nuonco/nuon/services/ctl-api/internal/app"
 	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/cctx"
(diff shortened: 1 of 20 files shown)


services/ctl-api/internal/app/installs/service/get_org_workflows.go
```diff
@@ -105,7 +106,9 @@ func (s *service) getOrgWorkflows(ctx *gin.Context, orgID string, excludePlanOnl
 		Preload("CreatedBy").
 		Preload("Steps").
```
ALPHAvibe: i would expect we need to limit or remove steps from the preloads. this has a nasty fan out with multiple workflows.

ALPHAvibe (dismissed):

ALPHAvibe (approved):

PR #1684, opened 2026-06-15
feat(ctl-api): enable runner job & log tail long polling by default

RealHarshThakur:
Defaults the `runner-job-long-poll` and `log-tail-long-poll` org feature flags to on for newly created orgs.


fidiego (approved):
LGTM

PR #1689, opened 2026-06-16
feat: (ctl-api) surface composite error when a component build has failed

onprem:
When a component deploy targets a build in a terminal failure state (error / policy_failed), freeze a typed `ComponentBuildUnavailableError` onto the `InstallDeploy` so the dashboard guides the user to rebuild the component, instead of showing a raw status string or hanging on polling.

- Add `ComponentBuildUnavailableError` composite error type (reason "failed"; "missing" reserved for a follow-up)
- Add `RecordDeployBuildUnavailableCompositeError` activity to record it onto `InstallDeploy.composite_error` (best-effort)
- `componentdeploysyncandplan` signal:
  - record the composite error when build polling fails on a terminal build state
  - treat `policy_failed` as terminal so the deploy fails fast instead of polling for up to an hour
  - prefer the latest active build over latest-any when resolving an empty BuildID, so a newer errored build can't mask an older deployable one
  - set the workflow step's deploy target before polling so the error is visible in the dashboard even when the build never becomes deployable

No frontend changes: the existing `CompositeError` renderer surfaces it on the deploy page. Follow-up will cover truly-missing builds via a WorkflowStep-level composite error.

Screenshot 2026-06-16 at 10 43 52 PM

### DIFF
--- a/services/ctl-api/internal/app/installs/deployerrors/component_build_unavailable_error.go
+++ b/services/ctl-api/internal/app/installs/deployerrors/component_build_unavailable_error.go
@@ -1,0 +1,105 @@
+package deployerrors
+
+import (
+	"fmt"
+
+	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/compositeerrors"
+)
+
+// ComponentBuildUnavailableErrorType is the discriminator for a deploy that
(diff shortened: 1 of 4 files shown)


jonmorehouse (approved):
This looks solid to me!

PR #1694, opened 2026-06-17
fix: app provisioning on gcp

jordan-acosta:
## Summary

Creating apps is failing in Nuon BYOC on GCP. The root cause is that it's trying to select an IAM role, even though it's on GCP. It's not switching on the platform. This PR:

- Adds a GAR repo response to provide the correct URL
- Adds a cloud platform switch to the app provision signal
- Validates the repo URL at config load time, to fail early instead of when you try to create an app the first time

Creating install stacks was also failing. Another case of not pivoting on the platform and still trying to use the AWS path. This PR includes a fix for this as well.


fidiego (dismissed):
LGTM

fidiego (approved):
LGTM

PR #1697, opened 2026-06-18
feat: unbounded preload for runner configs

ALPHAvibe:
this preload is unbounded on a query that can grow large on apps with many syncs. updating to use latest config.


jonmorehouse (approved):

fidiego (approved):
LGTM

PR #1709, opened 2026-06-22
fix(ctl-api): pivot app reprovision repo on cloud platform

someshkoli:
App reprovision still used the AWS ECR path for default orgs and failed on GCP/Azure control planes. #1694 fixed the provision signal but not reprovision (the GCP/Azure switch lived in the legacy `worker/reprovision.go` removed in #1532 and was never restored here).

This mirrors the provision signal: GAR on GCP, ACR on Azure, ECR on AWS, fake response for non-default orgs.


jordan-acosta (approved):

fidiego (approved):
LGTM

PR #1743, opened 2026-06-26
fix: handle helm tear downs where there is no previous release

jordan-acosta:
## Summary

While tearing down the karpenter_nodepools component in the Nuon BYOC on AWS, I found that the job will panic if it does not find a previous helm release. It looks like the chart is rendering an empty manifest in my test install. That looks like a config issue, but regardless the job should not panic.


fidiego (approved):
LGTM

PR #1772, opened 2026-07-01
fix: upsert gcp account in UpdateGCPAccountRegion if row missing

im-Amitto:
creates the gcp_accounts row on first phone-home when install was created without a pre-existing GCP account


im-Amitto:
moving to #1770

PR #1776, opened 2026-07-01
feat: change default for s3 backfill rate

ALPHAvibe:
change default from 100 to 500. this will improve speed when processing records already backfilled.


someshkoli (approved):

fidiego (approved):
LGTM

PR #1801, opened 2026-07-05
feat: experimental spacelift support

jordan-acosta:
## Summary

Driven by a potential customer interest in using Spacelift to manage install stacks, this PR adds support for provisioning an install stack using two new methods, along with some quality-of-life updates to the Terraform method.

The Spacelift method is behind a feature toggle.

### Spacelift Blueprints

Configure and provision an install stack using the Spacelift UI. The customer can install the blueprint, create a stack, and configure it using Spacelift stack inputs and secrets.

Screenshot 2026-07-05 at 13 42 37

### Spacelift Administrative Stacks

Configure and provision an install using the Spacelift Terraform provider. The customer can copy the generated terraform config into their VCS, and add the variables and secrets into Spacelift (or whatever secret store their using.)

Screenshot 2026-07-05 at 13 42 44

### Terraform

This PR also splits up the generated `install.tfvars` file to break out secrets, and renames the files so they will be picked up by the terraform CLI automatically.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -289,6 +289,7 @@
 				"icon": "clock-rotate-left",
 				"pages": [
 					"updates/updates",
+					"updates/037-split-install-stack-tfvars",
 					"updates/036-toggleable-components",
 					"updates/035-runbooks-and-readmes",
 					"updates/034-aws-terraform-install-stack",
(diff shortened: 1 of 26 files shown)


fidiego (approved):
shape LGTM
do we want to accept that specific version for `defaultSpaceliftTerraformVersion`?

PR #1806, opened 2026-07-06
fix: remove unwanted preloads and activites from hot path

someshkoli:
Get install and get full app config with full preloads are two of the heaviest and fattest queries we have, we query them even though we dont need full information about those entities, this pr trims it down where possible and skips on data passsing betwen activities and workflows to reduce overall temporal encoder do less work

### DIFF
--- a/services/ctl-api/internal/app/installs/worker/activities/action_workflow_in_app_config.go
+++ b/services/ctl-api/internal/app/installs/worker/activities/action_workflow_in_app_config.go
@@ -1,0 +1,32 @@
+package activities
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/nuonco/nuon/services/ctl-api/internal/app"
+)
+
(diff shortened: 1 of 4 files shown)


onprem (approved):

services/ctl-api/internal/app/installs/worker/activities/action_workflow_in_app_config.go
```diff
@@ -0,0 +1,32 @@
...
+
+	res := a.db.WithContext(ctx).
+		Model(&app.ActionWorkflowConfig{}).
+		Where(app.ActionWorkflowConfig{
+			AppConfigID:      req.AppConfigID,
+			ActionWorkflowID: req.ActionWorkflowID,
+		}).
+		Count(&count)
+	if res.Error != nil {
+		return false, errors.Wrap(res.Error, "unable to check action workflow in app config")
```
jonmorehouse, line 28: This needs to be wrapped with `generics.TemporalGormErr`

PR #1813, opened 2026-07-06
fix: dashboard ui failed deploy errors

ALPHAvibe:
Deploys (and any resource carrying a composite_error) fail to load in the dashboard with a "Refresh failed / failed to fetch deploys" toast. Root cause is a swagger type-generation bug: composite_error.data was generated in the nuon-go SDK as []int64, but the API returns it as an object, so the dashboard BFF's SDK decode blows up.


fidiego (approved):
LGTM

PR #1829, opened 2026-07-07
fix: admin request still process on missing admin account

ALPHAvibe:
fix for admin endoints still processing if an invalid email is passed into the header request.


fidiego (approved):
LGTM

PR #1830, opened 2026-07-07
fix:  control-plane component builds ecr auth 

RealHarshThakur:
Wires the OCI resolver into the container image handler, assumes the management role for control-plane component and sandbox build registry auth, and writes step failures into the build log stream.


fidiego (approved):
LGTM

PR #1844, opened 2026-07-08
feat: (ctl-api) helm tool-layer composite-error parser

onprem:
Add a `LayerTool` parser for helm jobs. This parser would try to parse all helm related errors in component deploys and store them as composite errors.

The runner drives helm via the Go SDK, so it anchors on the runner's error wrapper (unable to ... helm release: / with dry-run:) to strip nesting and lead with the real SDK cause, with verified helm v4 SDK cause strings as a backup anchor. Generic kubernetes phrases are excluded as signals so streamed pod-log lines can't hijack the headline.


fidiego (approved):
LGTM

PR #1845, opened 2026-07-08
fix: show builds for orgs that build on the control plane

RealHarshThakur:
Orgs that build on Nuon's control plane have no org runner, which left
several gaps in the dashboard and the org lifecycle. This closes them.

**The builds page lists control-plane build jobs.** Orgs with
control-plane builds and no org runner used to see an empty "no build
executor configured" state. They now get a build jobs timeline under a
header that explains the mode.

**Disabling control-plane builds restores the org runner.** Turning the
feature off re-enables the org runner, and reprovisioning the org creates
the missing runner group instead of completing with no effect.

**Admin controls hide the empty org runner ID.** The Org Runner ID card no
longer renders a blank box for orgs that have no runner.

## What stays the same

Orgs that already have an org runner are unaffected: same builds page, same
reprovision behavior. Enabling control-plane builds does not tear down an
existing runner; that remains a deliberate action rather than a side effect
of the toggle.

## Mechanics worth flagging

Control-plane builds and the org runner are kept mutually exclusive:
changing one derives the other from the org's current state. Provision and
reprovision now create the org runner group when it is missing and pass the
runner identity into IAM provisioning, so GCP and Azure workload identity
are set up correctly for a freshly created runner.


fidiego (approved):
LGTM

PR #1875, opened 2026-07-11
fix: fetch app config schemas from their own urls

jordan-acosta:
## TLDR

Some TOML plugins ignore the `type` query param when caching schemas. Even with the recent UID fix, these LSPs will use the same schema (whichever they download first,) for all files. This PR fixes that by adding a `/config-schema/:type` handler alongside `/config-schema?type=:type`.

> [!NOTE]
> The previous endpoint will keep working as it does today, but some LSPs will require migrating app configs to the new endpoint.

## Summary

This appears to be a divergence between the two most popular TOML LSPs: Taplo and Tombi. Taplo is older, more popular, and used by plugins like VSCode's "Even Better TOML". Tombi is newer, but growing, and is used in Zed and Neovim plugins like coc-toml.

It looks like Tombi strips the query params when caching the schemas. This causes it to use whichever schema it fetches first for all files. The fix is to simply give each type it's own URL, which is used to generate the type ID.

Additionally, the domain is included in the ID. This should avoid conflicts when working with multiple instances of Nuon on different versions.


fidiego (dismissed):
LGTM

fidiego (approved):
LGMT

PR #1888, opened 2026-07-13
fix: Accept nil env vars in shell hooks

### LINKED ISSUE #1887: Runner >= 0.19.1054: sandbox runs fail with "unable to get hooks: EnvVars required" when sandbox env vars are empty
**Versions:** runner 0.19.1054 and 0.19.1055 (0.19.1050 unaffected), Nuon SaaS control plane (api.nuon.co), CLI 0.19.899.

**Repro:**
1. App with `runner_type = "azure"` and a terraform sandbox config with no `env_vars`.
2. Runner on >= 0.19.1054.
3. Trigger any sandbox run (e.g. reprovision).

**Expected:** the run executes terraform.

**Actual:** every sandbox run fails at workspace creation, before terraform runs:

```
execute: unable to execute job: unable to execute apply-plan run: unable to execute apply pipeline:
unable to execute step 8.apply plan: unable to execute: error running apply: exit status 1
unable to create workspace from config: unable to get hooks:
Key: 'shell.EnvVars' Error:Field validation for 'EnvVars' failed on the 'required' tag
```

**Cause:** 0.19.1054 switched the sandbox plan fetch from `GetJobPlanJSON` to `GetJobCompositePlan` (`bins/runner/internal/jobs/sandbox/terraform/fetch.go`). `CompositePlanFromAny` converts the SDK model via a JSON round-trip, and the generated model field `EnvVars map[string]string \`json:"envVars,omitempty"\`` drops an empty map during re-marshal. `Hooks.EnvVars` therefore arrives nil, and `shell.New` (`pkg/terraform/hooks/shell/shell.go`) rejects it via `validate:"required"`.

The empty-map case is routine: `getSandboxRunEnvVars` (ctl-api) injects a region var for AWS and GCP runners but nothing for Azure, so any Azure-runner app without explicit sandbox `env_vars` hits this on every sandbox run. The API-side composite plan is fine (`"EnvVars": {}`); the value is lost client-side in the round-trip — same class of artifact as the empty `aws_auth` case already handled by `pruneEmptyClusterAuth` in `CompositePlanFromAny`.

Nil EnvVars would also panic later if validation passed: `setCredentials` merges into `s.EnvVars` via `generics.MergeMap`, which writes into the map.

**Workaround:** set any entry under `env_vars` in the app sandbox config.

Fix PR incoming: accept nil env vars in shell hooks and normalize to an empty map.

ehe3:
Fixes #1887.

Since 0.19.1054, every sandbox terraform run fails at workspace creation with `unable to get hooks: Key: 'shell.EnvVars' ... 'required'` when the plan's hook env-var map is empty — which is every Azure-runner app without explicit sandbox `env_vars`, since only AWS/GCP runners get a region var injected. The map is present in the API-side composite plan (`"EnvVars": {}`) but is dropped by `omitempty` during the SDK-model JSON round-trip in `CompositePlanFromAny`, so it reaches `shell.New` as nil and fails the `required` validation. Full trace in the issue.

Hooks with no env vars are a valid state, so this drops the `required` tag and normalizes nil to an empty map in `New`. The normalization also prevents a latent panic: `setCredentials` merges fetched credentials into `s.EnvVars` via `generics.MergeMap`, which writes into the map, so a nil map would blow up at hook execution even without the validator.

Verified against a real install: an Azure sandbox run failing with this error succeeds once the env-var map is non-empty (config workaround), and the added tests cover the nil and populated cases.

An alternative (or complement) would be normalizing `SandboxRunPlan.Hooks.EnvVars` in `CompositePlanFromAny`, alongside the existing `pruneEmptyClusterAuth` handling of the same round-trip artifact — happy to move the fix there if you prefer.


jordan-acosta:
@ehe3 , this fix looks good, but we require signed commits to merge. Could you sign your commit and update this PR? Once updated the PR checks should pass.

ehe3:
> @ehe3 , this fix looks good, but we require signed commits to merge. Could you sign your commit and update this PR? Once updated the PR checks should pass.

done!

jordan-acosta:
Thanks, @ehe3 . We can merge as soon as the PR checks pass.

jordan-acosta:
@ehe3 , due to the way our PR checks are implemented, they cannot run against a PR from a fork. I took the liberty of opening my own PR to merge your changes: https://github.com/nuonco/nuon/pull/1919

Apologies for the confusion here. We'll update our contribution guide to make this clear.

PR #1919, opened 2026-07-15
fix: Accept nil env vars in shell hooks

jordan-acosta:
Sandbox run plans can have hooks enabled with no env vars: Azure runner apps get no injected vars (AWS/GCP always get a region var), and an empty env-var map is dropped by omitempty during the composite-plan JSON round-trip introduced in 0.19.1054, so it reaches the runner as nil. The required validation then fails every sandbox terraform run with 'unable to get hooks' before terraform executes.

Drop the required tag and normalize nil to an empty map so setCredentials can merge into it.

### DIFF
--- a/pkg/terraform/hooks/shell/shell.go
+++ b/pkg/terraform/hooks/shell/shell.go
@@ -23,7 +23,7 @@ type shell struct {
 	v       *validator.Validate
 	rootDir string
 
-	EnvVars map[string]string `validate:"required"`
+	EnvVars map[string]string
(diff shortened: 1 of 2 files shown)


jordan-acosta:
Our PRs checks won't run against PRs from forks, so this PR can't be merged: https://github.com/nuonco/nuon/pull/1888

Opening this PR to merge the the change while we decide how we want to handle this long-term.

nnnnat (approved):

PR #1934, opened 2026-07-17
feat: `nuon installs` sub-command hierarchy and bugs

jordan-acosta:
## Summary

While adding a new sub-command under `nuon installs`, I noticed that our command hierarchy was inconsistent. This made knowing where to add the sub-command ambiguous, and users might not know where to look for it. I wanted to take a pass at making `nuon installs` consistent. If we like this direction, we can update the rest of the CLI to follow.

While working on this, I also found that all commands do not return output in a consistent way, so this PR includes fixes for those as well.

### Sub-command hierarchy

Basically, there were two ways sub-commands were organized. They could be under `nuon installs` directly with a compound name, like `nuon installs teardown-component`. Or they could be nested under a sub-command, like `nuon components toggle`. I decided to try nesting all the sub-commands consistently, then grouping the remaining commands.

```sh
  Manage installs

  USAGE

    nuon installs [command] [--flags]

  INSTALL COMMANDS

    create [--flags]                Create an install
    delete [--flags]                Delete install
    deprovision [--flags]           Deprovision install
    deselect                        Deselect your current install
    forget [--flags]                Forget install
    get [--flags]                   Get an install
    list [--flags]                  List installs
    reprovision [--flags]           Reprovision install
    select [--flags]                Select your current install

  INSTALL CONFIG COMMANDS

    generate-config [--flags]       Generate config for an existing install
    sync [--flags]                  Sync install
    toggle-sync [--flags]           Enable/Disable install config sync

  INSTALL CHILD COMMANDS

    actions [command] [--flags]     Manage install actions [preview]
    components [command] [--flags]  Manage install components
    deploys [command]               Manage install component deploys
    inputs [command] [--flags]      Manage install inputs
    labels [command]                List, set, or unset labels on an install
    outputs [--flags]               View install outputs
    runner [command]                Manage install runner
    sandbox [command]               Manage an install sandbox
    sandbox-runs [command]          Manage install sandbox runs
    stacks [command]                View install stacks
    workflows [command] [--flags]   Manage workflows
```

### Command aliases

In some cases, making things consistent resulted in commands that were kind of long and annoying to type. I like the consistency, and it's probably better for agents, but I still want humans to enjoy using this, so I added a few aliases for convenience. For example:

```sh
# the "real" command
nuon installs sandbox-runs create --type=reprovision
# a more intuitive alias
nuon installs sandbox reprovision
```

### Response formatting

Commands were not returning consistently formatted responses. A few examples:

- Some `outputs` commands return a table, where other print JSON.
- Some `create` commands display a TUI, others static text.
- Some `logs` commands open a URL in the browser, other display a TUI.

This PR resolves these inconsistencies as well.


erickyellott (approved):
🔥

PR #1942, opened 2026-07-18
fix: cors swagger

ALPHAvibe:
Mixed Content: The page at https://app.nuon.co/admin/swagger/... was loaded over HTTPS, but requested an insecure resource http://ctl.nuon.us-west-2.prod.internal.nuon.co/v1/installs.... This request has been blocked.

Root cause

Swagger UI builds its Execute request from the spec's own host / schemes / basePath, not from the page. The ctl-api admin spec sets host from AdminAPIURL (the internal host) with schemes: ["http"], so Execute fires at http://ctl...internal.nuon.co/v1/installs — a cross-origin, insecure, internal-only URL. On an HTTPS page that's blocked as mixed content before the request is even sent (and the host isn't browser-routable anyway).

The dashboard BFF already proxies the Swagger UI, but newSwaggerProxy only rewrote the docs HTML (the embedded spec URL). The spec JSON passed through untouched, so the internal host stayed baked in. This never surfaced locally because dev is all http (no mixed-content rule) and ctl-api's CORS allows all origins — so "CORS" in the Swagger boilerplate was a red herring; mixed content was the actual blocker.

Fix

Extend newSwaggerProxy's ModifyResponse to also rewrite the spec JSON (/oapi/v2 and /oapi/v3) so Execute stays same-origin and routes back through the BFF:

- Swagger 2.0: drop host and schemes (Swagger UI falls back to the page origin + scheme) and set basePath to the BFF prefix.
- OpenAPI 3: rewrite servers to the same prefix.
- The prefix is per-surface: /admin for the admin proxy (matches the existing /admin/v1/* route), "" for public (matches /v1/*).


nnnnat (dismissed):

fidiego (approved):
LGTM

erickyellott (approved):

PR #1946, opened 2026-07-19
feat: azure per-operation runner managed identities

im-Amitto:
Adds Azure per-operation user-assigned managed identities so the install runner assumes a scoped identity per operation (provision/maintenance/deprovision/custom/break-glass) instead of holding broad standing grants — the Azure equivalent of AWS assume-role / GCP impersonation. When Azure roles are configured, the broad grants (Contributor/RBAC-Admin/AKS-Admin/register-action) are stripped from the runner's VMSS system identity; installs without Azure roles keep legacy behavior. App config gains `azure_actions` / `azure_built_in_roles`, and the ARM template creates + attaches the identities and surfaces their client IDs via phone-home.


pkg/kube/cluster_info.go
```diff
@@ -167,18 +167,23 @@ func ConfigForCluster(ctx context.Context, cInfo *ClusterInfo) (*rest.Config, er
 		if os.Getenv("ARM_USE_MSI") == "true" {
 			loginMode = "msi"
 		}
+		args := []string{
+			"get-token",
+			"--login",
+			loginMode,
+			"--server-id",
+			"6dae42f8-4368-4678-94ff-3960e28e3630",
```
erickyellott, line 175: This is hardcoded?

erickyellott (approved):

pkg/kube/cluster_info.go
im-Amitto: it's azure k8s AAD Server application ID, constant

PR #1951, opened 2026-07-20
fix(components): show source failures in build history

RealHarshThakur:
## Summary

Component builds now exist before Nuon checks their connected repository source. Invalid branches and other source lookup failures appear as failed builds with logs instead of leaving app sync waiting for a build record that never arrives.

**See source lookup failures in component build history.** Nuon creates the build and its log stream before resolving the configured branch. A failed lookup marks the build as failed and stops before creating a runner job.

**Retry against a stable commit.** Successful source resolution records the commit SHA on the build. Planning and retries reuse that SHA instead of resolving the branch again.

**Get consistent behavior from app sync and app builds.** Both queued sync builds and app-level build workflows use the same preflight ordering.

## What stays the same

Manual component builds retain their existing eager source validation. Public repositories, non-Git components, and explicit branch-run commit pins behave as before.

### DIFF
--- a/services/ctl-api/internal/app/apps/signals/branches/activities/create_component_build.go
+++ b/services/ctl-api/internal/app/apps/signals/branches/activities/create_component_build.go
@@ -36,7 +36,7 @@ func (a *Activities) CreateComponentBuild(ctx context.Context, req CreateCompone
 	ctx = cctx.SetOrgIDContext(ctx, cmp.OrgID)
 	ctx = cctx.SetAccountIDContext(ctx, cmp.CreatedByID)
 
-	build, err := a.componentHelpers.CreateComponentBuildWithID(ctx, req.BuildID, req.ComponentID, true, nil)
+	build, err := a.componentHelpers.CreateComponentBuildWithID(ctx, req.BuildID, req.ComponentID, false, nil)
 	if err != nil {
 		res := a.db.WithContext(ctx).Where(&app.ComponentBuild{ID: req.BuildID}).First(&existing)
 		if res.Error == nil {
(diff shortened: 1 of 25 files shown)


services/ctl-api/internal/app/runners/service/list_control_plane_build_activity.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// @ID ListControlPlaneBuildActivity
+// @Summary list control-plane build activity
+// @Tags runners
+// @Produce json
+// @Security APIKey
+// @Security OrgID
+// @Param offset query int false "offset of activity to return" Default(0)
+// @Param limit query int false "limit of activity to return" Default(10)
```
ALPHAvibe: missing page query param

https://github.com/nuonco/nuon/blob/a2410397ca7a7387533df8a3be79bdae491fc818/services/ctl-api/internal/app/orgs/service/get_all_orgs.go#L23

services/ctl-api/internal/app/runners/service/list_control_plane_build_activity.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	offset, err := strconv.Atoi(ctx.DefaultQuery("offset", "0"))
+	if err != nil || offset < 0 {
+		ctx.Error(stderr.ErrUser{Err: fmt.Errorf("invalid offset %q", ctx.Query("offset")), Description: "invalid offset"})
+		return
+	}
+
+	candidates := []controlPlaneBuildActivityCandidate{}
+	res := s.db.WithContext(ctx).Raw(`
+		SELECT kind, id, created_at FROM (
```
ALPHAvibe: pagination will not play nice here with the union. I would recommend splitting this up into two paginations to fetch the date you need here and merge what you need on the client.

ALPHAvibe (approved):

PR #1954, opened 2026-07-20
feat: apply install stacks through AWS connections

RealHarshThakur:
## Summary

Organizations can connect an AWS account and let Nuon apply the initial install stack through a customer-managed cross-account role. The capability is disabled by default and must be enabled per organization.

**Connect and verify an AWS account.** Users provide an account ID, region, and customer role ARN. Nuon generates an External ID and trust policy, then verifies the complete management-role-to-customer-role chain.

**Re-check connection access.** Users can verify a connection again after its role or trust policy changes. Connection lists do not expose External IDs or trust policy details.

**Choose who applies an install stack.** AWS install forms default to customer-managed application. Selecting a verified connection lets Nuon create the initial CloudFormation stack and continue through the existing phone-home workflow.

## What stays the same

Installs without a connection continue through the existing manual stack flow. Sandbox, Azure, GCP, AWS-EKS, AWS-ECS, reprovision, and app-update workflows do not invoke managed CloudFormation creation.

## Mechanics worth flagging for review

Nuon first assumes the existing management role, then assumes the customer role with its External ID. Temporal carries only install, stack-version, and connection IDs. CloudFormation retries reconcile against the stack operation token, and connections used by installs cannot be deleted.


fidiego (commented):
lgtm

add a test plan?

fidiego (approved):

PR #1982, opened 2026-07-28
fix: retry failed migrations instead of skipping them

im-Amitto:
Hit this upgrading a BYOC install. Two backfills failed, and the deploy after them went green with both never applied.

## what happened

`isMigrationApplied` only checked that a row existed, not its status. So when 113 errored, the next run read the error row as applied, skipped it, and moved on to 114. Same again with 116. Each deploy quietly advanced past one broken migration:

- 09:25 ran 109-112, died on 113
- 09:28 skipped 113, ran 114-115, died on 116
- next run skipped 116, ran 117-118, exit 0

So a single failure was permanent and invisible. `inspect_migrations` showed both sitting at `status=error` while the deploy reported success.

## why 113 and 116 failed

Same root cause. Both create records through models, and those BeforeCreate hooks pull `created_by_id` off the request context:

```go
if i.CreatedByID == "" {
    i.CreatedByID = createdByIDFromContext(tx.Statement.Context)
}
```

A migration runs on `context.Background()`, so there is no account, `created_by_id` goes in as `''`, and `created_by_id` has an FK to accounts. Every insert fails. These only ever worked where the loop body iterated zero rows, which is why they passed in an env with no pre-existing runners or installs to backfill and failed on the first install that actually had data.

Added `systemCtx` to attribute migration writes to the `nuonctl` service account, which 01-create-internal-accounts already creates.

Two more on top of that:
- 116 left `OldAppConfigID` empty, and `''` fails the app_configs FK too. Backfilled rows have no previous config, so it needs to be NULL.
- 113 returned on the first bad runner, which aborted the backfill for every remaining one. Logs and continues now.

## also

`DB_MIGRATIONS_TIMEOUT` overrides the hardcoded 5m. A `CREATE INDEX` on queue_signals blew past it on this install, and since the deploy action's own timeout is longer, the 5m was the real ceiling and nothing outside the image could change it.

## notes

- `claimMigration` reclaims error rows always, and in_progress rows only after an hour, so a genuinely concurrent run still gets skipped rather than double-running.
- Existing error rows need deleting to re-run: `DELETE FROM migrations WHERE status = 'error'`. Worth checking installs for older poisoned rows, they have been silently skipped up to now.
- Separate follow-up: `applyIndex` emits a plain `CREATE INDEX`, which locks out writers on hot tables. Switching to CONCURRENTLY needs an invalid-index sweep first, because `HasIndex` queries `pg_indexes` and would treat a failed concurrent build as present.


im-Amitto:
reopening from am/migration-error-retry, wrong branch prefix and stale base

PR #1986, opened 2026-07-28
feat: component health

im-Amitto:
Adds live health as a fourth status axis next to deploy status. The runner watches each component's cluster resources and runs vendor-declared http/tcp/exec probes; the server debounces those observations into a per-component verdict (three bad reports to go bad, two to recover) that shows up in the dashboard, `nuon installs health`, and Slack/webhook alerts. Components can opt into `block_deploy` to hold a deploy until health stabilizes, and dependency-aware suppression means one root cause produces one alert instead of one per dependent.

All of it sits behind the `component_health` org feature flag. The rule the whole design follows is that health never claims more confidence than it has: `unknown` means absence of information, so it never outranks a resource that did report and never alerts. Tested end to end against a real EKS install — that testing is where most of the interesting bugs came from, including a component reporting healthy off a passing probe while the runner had lost cluster access entirely.


someshkoli (dismissed):

im-Amitto:
@copilot resolve the merge conflicts in this pull request

someshkoli (approved):

PR #2048, opened 2026-08-03
feat: server-side app config sync

im-Amitto:
Removes `pkg/config/sync/apisyncer`. `nuon apps sync` now pushes the parsed config to the API in intermediate form and calls the new `POST /v1/apps/:app_id/configs/:config_id/sync`, which runs the shared DB syncer on the app queue. The syncer picks up the pieces that lived client-side: build scheduling with checksum reuse, install rollout, journey step, and orphan reporting via `state.result`. Follow-up to #2047.


someshkoli (dismissed):

fidiego (approved):
LGTM

PR #2065, opened 2026-08-03
chore: migrate pending process healthcheck emitter queue to cron namespace and task queue

someshkoli:
In earlier change set we moved all the crons to their dedicated namespace and workers, this change continues on that and moves / migrated pending process healthcheck queues, emitters and other parts of it to the new namespace, task queue and workers.


RealHarshThakur (approved):

im-Amitto (approved):

someshkoli:
not needed anymore

PR #2069, opened 2026-08-03
feat: run workflows on a resident host with strict cancel/retry semantics

RealHarshThakur:
## Summary

New workflows now run on a single resident workflow host, and step outcomes are decided by the workflow's actual result instead of signal delivery. Cancelled and failed steps stop the workflow instead of letting it carry on.

## What changes

**Every new workflow runs as a resident host.** Retries, skips, cancels, and appended steps land on the same live workflow instead of spawning per-retry clones. Idle hosts unwind after 15 minutes and re-warm on demand when you act on them.

**Cancelling a step or workflow stops it for good.** The next step never starts, cancellation shows up as cancelled (not success, not a retryable error), and a late success write can't overwrite it.

**A failed step only continues when it's explicitly marked skippable.** Non-skippable failures and unrecognized outcomes fail closed and stop the group.

**Retrying an apply step re-runs plan and apply together.** Action steps keep retrying on their own.

**When automatic retries run out, the step parks and waits for a manual retry** instead of failing the whole workflow. Completion notifications (webhooks, Slack) hold until the workflow actually finishes.

**Polling a finished workflow no longer restarts it.** Dashboard status checks on a terminal flow are read-only.

## What stays the same

Workflows already in flight before the deploy keep the old execution paths. Superseded retry attempts keep their errors visible in run history.
## Tests

Property tests pin the dispatch contract: cancelled, failed non-skippable, and unknown outcomes never dispatch the next step; a retried apply always re-plans. Full integration suite passed clean (41/41) on the final code.


services/ctl-api/internal/app/actions/service/create_adhoc_action.go
```diff
@@ -162,9 +162,7 @@ func (s *service) CreateAdHocAction(ctx *gin.Context) {
 		ctx.Error(err)
 		return
 	}
-	if err := s.enqueueInstallSignal(ctx, queueID, &executeflow.Signal{
-		WorkflowID: workflow.ID,
-	}, workflow.ID, "install_workflows"); err != nil {
+	if err := s.enqueueInstallSignal(ctx, queueID, executeflow.NewSignal(workflow.ID), workflow.ID, "install_workflows"); err != nil {
```
someshkoli, line 165: remove hardcoded names

services/ctl-api/internal/app/installs/service/create_install.go
```diff
@@ -122,9 +122,7 @@ func (s *service) CreateInstallV2(ctx *gin.Context) {
 		ctx.Error(fmt.Errorf("enqueue signal: %w", err))
 		return
 	}
-	if err := s.enqueueInstallSignal(ctx, workflowsQueueID, &executeflow.Signal{
-		WorkflowID: workflow.ID,
-	}, workflow.ID, "install_workflows"); err != nil {
+	if err := s.enqueueInstallSignal(ctx, workflowsQueueID, executeflow.NewSignal(workflow.ID), workflow.ID, "install_workflows"); err != nil {
```
someshkoli, line 125: same ^

services/ctl-api/internal/app/runners/service/admin_upsert_sandbox_config.go
```diff
@@ -35,7 +36,11 @@ func (s *service) AdminUpsertSandboxConfig(ctx *gin.Context) {
 		return
 	}
 
-	if req.JobType == "" {
+	jobType := ctx.Param("job_type")
+	if jobType == "" {
+		jobType = req.JobType
+	}
+	if jobType == "" {
```
someshkoli, line 43: why is this even needed, why send same thing via body and param ?

services/ctl-api/internal/app/runners/service/admin_upsert_sandbox_config.go
```diff
@@ -64,7 +70,7 @@ func (s *service) AdminUpsertSandboxConfig(ctx *gin.Context) {
 	if res := s.db.WithContext(ctx).
 		Clauses(clause.OnConflict{
 			Columns:   []clause.Column{{Name: "job_type"}, {Name: "operation"}, {Name: "deleted_at"}},
-			DoUpdates: clause.AssignmentColumns([]string{"enabled", "duration", "sleep_duration", "should_error", "panic", "trigger_shutdown", "log_template", "plan_template", "plan_display_template", "state_template", "output_template", "updated_at"}),
+			DoUpdates: clause.AssignmentColumns([]string{"enabled", "duration", "sleep_duration", "should_error", "error_message", "panic", "trigger_shutdown", "log_template", "plan_template", "plan_display_template", "state_template", "output_template", "updated_at"}),
```
someshkoli, line 73: footgun with hardcoded col names

services/ctl-api/internal/pkg/callback/await.go
```diff
@@ -59,10 +77,27 @@ func AwaitWithTimeout(ctx workflow.Context, ref Ref, timeout time.Duration) (*Re
...
+		case "error":
+			msg := result.StatusDescription
+			if msg == "" {
+				msg = "failed without further detail"
+			}
 			return nil, temporal.NewNonRetryableApplicationError(
-				result.StatusDescription,
+				msg,
 				"SIGNAL_FAILED", nil)
+		case "cancelled":
```
someshkoli, line 93: enums please

services/ctl-api/internal/pkg/flow/signals/executeflow/execute.go
```diff
@@ -625,6 +691,10 @@ func (s *Signal) handle(ctx workflow.Context, startFromGroupIdx int) error {
 		}
 	}
 
+	if !s.isWorkflowComplete(ctx) {
```
someshkoli, line 694: this will fail if any step is in error state and workflow is continued

services/ctl-api/internal/pkg/flow/signals/executeflow/execute.go
```diff
@@ -706,16 +776,16 @@ func (s *Signal) findGroupPositionForStep(ctx workflow.Context, stepID string) i
...
 	if err != nil || len(groups) == 0 {
-		return 0, false
+		return residentScheduleDecision{State: residentScheduleRunnable}
 	}
 	steps, err := workflowactivities.AwaitPkgWorkflowsFlowGetFlowSteps(ctx, workflowactivities.GetFlowStepsRequest{
 		FlowID: s.WorkflowID,
 	})
 	if err != nil {
-		return 0, false
+		return residentScheduleDecision{State: residentScheduleComplete}
```
someshkoli, line 788: if err nil then workflow completed ? 

Image

someshkoli (changes requested):

PR #2103, opened 2026-08-06
fix: stop shipping the runner api token to customer accounts

im-Amitto:
We were putting the runner API token in two places a customer can read it, and nothing consumes it — runners bootstrap through `/v1/runner-auth/{aws,aws-iid,gcp,azure}`, and the init script only reads `nuon_runner_api_url` and `nuon_runner_id` from tags before calling `mng fetch-token`.

First, we set it as a CloudFormation **stack tag**, which propagates to every taggable resource in the stack. Dropped. Second, the onboarding builder still pinned the nested templates to `v0.2.1` — the last version whose runner template declares `RunnerApiToken` (no `NoEcho`, so visible in DescribeStacks) and writes it into the launch template's instance tags. The templates repo removed all of that in `v0.2.2` back in March; we just never bumped. Now on `v0.4.2`.

The bump also drops the nested template's `InstanceType` `AllowedValues`, which is what makes `t3a.medium` unusable in eu-north-1 — so this should close the InstanceType half of Pylon #2087 as well as #1927. Safe on parameters: the only one removed between v0.2.1 and v0.4.2 is `RunnerApiToken`, which we already send conditionally on the template declaring it, and the two VPC additions (`EnableFirewall`, `EgressAllowedDomains`) both default off.


im-Amitto:
Evidence for the template-repo half, from the published artifact `aws-cloudformation-templates/v0.1.6/runner/asg/stack.yaml`:

```yaml
# line 14 — no NoEcho, so the value is visible in DescribeStacks / console / stack events
  RunnerApiToken:
    Type: String
    Description: API token for the runner

# line 178 — written into the launch template's instance TagSpecifications
          - Key: nuon_runner_api_token
            Value:
              Ref: RunnerApiToken
```

`grep -c NoEcho` on that template returns 0.

Two changes needed there:
1. `NoEcho: true` on `RunnerApiToken` — one line, no runtime impact.
2. Drop the `nuon_runner_api_token` instance tag. Safe on the current init script (`scripts/aws/init-mng.sh` reads only `nuon_runner_api_url` + `nuon_runner_id`, then `mng fetch-token` with `RUNNER_AUTH_METHOD=iid`), but worth checking whether any customer pins an older `init_script_url` that reads the token from tags — the seed tracks `refs/heads/main`, so most follow main.

Longer term the parameter can go away entirely, since all four auth methods are keyless.

PR #2104, opened 2026-08-06
fix: release emitters blocked by expired signals

im-Amitto:
A cron emitter refuses to fire while it has a signal in `queued`/`in-progress`, and signal expiry is only ever enforced *inside* the handler (`handler/state.go:17` at init, and the `AliveChecker` in `handler/run.go:125`). So when a handler dies — worker OOM, terminate, continue-as-new mid-execute — its row stays in-flight forever and that emitter never fires again. The cron just silently stops.

`MaxInFlightAge` already covers this, but only four signal types declare it, and the two that don't are drift checks and cron-triggered actions — both of which set `SignalExpiresIn: 15m` and so had no external escape at all.

`EmitSignal` now treats a past-`ExpiresAt` signal as no longer holding the emitter, marking it errored the same way the max-age path does, with `stale_drop` recording which rule fired. This is a no-op when the handler is alive: it would hit the same expiry itself and set the row to error via `WithOnStopped`, releasing the emitter anyway. It only changes the case where nothing is left to do that. The decision moved into `partitionStaleSignals` so it's testable without standing up Temporal.

### DIFF
--- a/services/ctl-api/internal/app/apps/signals/branches/builds/execute_test.go
+++ b/services/ctl-api/internal/app/apps/signals/branches/builds/execute_test.go
@@ -3,10 +3,12 @@ package builds
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	"go.temporal.io/sdk/testsuite"
+	"go.temporal.io/sdk/worker"
 	"go.temporal.io/sdk/workflow"
 
 	"github.com/nuonco/nuon/services/ctl-api/internal/app"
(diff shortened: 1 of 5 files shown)


RealHarshThakur (dismissed):

im-Amitto:
CI failure was `[TMPRL1101] Potential deadlock detected: workflow goroutine "root" didn't yield for over a second` in `branches/builds` — the Temporal test-env watchdog firing under a loaded runner, in a package this PR doesn't touch (it also failed the same way on #2101).

Rather than just re-running, fixed the flake: three test files build a test workflow environment without raising `DeadlockDetectionTimeout` off its 1s default, while `planinstallgroup` and `appconfig` already set it to a minute — and those two passed both runs. Now consistent across all five.

Verified the option is load-bearing rather than assuming: setting it to `time.Millisecond` reproduces the exact CI error locally, `time.Minute` passes.

PR #2128, opened 2026-08-10
feat: denormalize install/app ancestry onto polymorphic-owner tables

jordan-acosta:
Adds nullable install_id/app_id columns to install_workflows, runner_jobs, queues, log_streams, and terraform_workspaces. Both NULL means org-tier, a defined state. Columns are populated in BeforeCreate via app.ResolveOwnerAncestry and backfilled by migration 126, batched and idempotent, one targeted update per (table, owner-type) pair.

A passive schema change: nothing reads the columns yet. The authorization enforcement and per-account grant layer that consume them live on ja/account-resource-grants, which stacks on this branch. Design docs in specs/auth.

### DIFF
--- a/sdks/nuon-go/models/app_log_stream.go
+++ b/sdks/nuon-go/models/app_log_stream.go
@@ -17,6 +17,9 @@ import (
 // swagger:model app.LogStream
 type AppLogStream struct {
 
+	// app id
+	AppID string `json:"app_id,omitempty"`
+
 	// attrs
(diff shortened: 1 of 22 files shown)


jordan-acosta:
Deferring: nothing merging now reads the denorm columns — they're prerequisite for the next phase (per-install grants / nested-route enforcement). Branch ja/resource-grants-v2 stays; will reopen or recreate when that work starts.

PR #2130, opened 2026-08-10
feat: CLI + docs for access control, plus role-reconcile perf

jordan-acosta:
### Summary

Follow-ups to the roles cleanup in #2123. We decided **not** to add the domain-scoped App developer / Install manager roles, so this PR no longer touches the role catalog or authorization — it's the surrounding polish that stands on its own.

**1. `nuon roles list` shows only assignable roles**

Held-only roles carry no assignment contexts, so the CLI now filters out roles with an empty `applies_to` — dropping the deprecated `org_support`/`installer` and the machine-only `runner` from the output. Matches the command's purpose of listing roles you can assign to members and service accounts. Filtered client-side so the unfiltered `GET /v1/roles` (used by the dashboard to show titles for already-held roles) is unaffected.

**2. Access control docs**

There was no single place documenting control-plane roles, and the scattered mentions were stale/inconsistent. This adds a dedicated **Access Control** concept page (Admin, Read-only; how roles are assigned across team invites, service accounts, API tokens, and OIDC trust policies; that tokens/trust-policies are backed by a service account carrying the role) and points the concept pages, CLI reference, dashboard, and security docs at it. Also fills the team-management guide's gap (assigning a role on invite, changing a member's role).

**3. Reconcile performance**

Independent improvements to `ReconcileOrgRoles` (called at org creation and by migration 126):
- Add a `roles(org_id, role_type)` index so the reconcile `SELECT` and the runtime `GET /v1/roles` use an index scan instead of a sequential scan.
- Load an org's roles in one query, skip metadata updates that already match, batch missing-role creates, and run writes in a single transaction.

### Testing

- `authz`, `authz/permissions`, `middlewares/org`, and `migrations` packages build and pass; CLI builds.
- Docs: no stale role references; `docs.json` valid.

### DIFF
--- a/bins/cli/internal/services/roles/list.go
+++ b/bins/cli/internal/services/roles/list.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/nuonco/nuon/bins/cli/internal/ui"
+	"github.com/nuonco/nuon/sdks/nuon-go/models"
 )
 
 func (s *Service) ListRoles(ctx context.Context, asJSON bool) error {
@@ -15,9 +16,18 @@ func (s *Service) ListRoles(ctx context.Context, asJSON bool) error {
 
 	view := ui.NewListView()
 
-	roles, err := s.api.ListRoles(ctx)
+	allRoles, err := s.api.ListRoles(ctx)
 	if err != nil {
 		return view.Error(err)
+	}
+
+	// Only list roles that can be assigned somewhere; held-only roles
+	// (deprecated or machine-only) carry no assignment contexts.
+	roles := make([]*models.AppRole, 0, len(allRoles))
+	for _, r := range allRoles {
+		if len(r.AppliesTo) > 0 {
+			roles = append(roles, r)
+		}
 	}
 
 	if asJSON {
--- a/docs/cli-commands.mdx
+++ b/docs/cli-commands.mdx
@@ -31,10 +31,10 @@ These work on every command.
 |---------|--------------|
 | `nuon orgs select` | Switch your active org. A new trial account has a single org that's selected automatically when you log in, so you only need this once you belong to more than one. |
 | `nuon orgs current` | Show the active org. |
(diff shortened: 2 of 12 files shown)


services/ctl-api/internal/pkg/db/psql/migrations/127_backfill_domain_roles.go
```diff
@@ -0,0 +1,37 @@
...
+		var orgs []app.Org
+		if err := db.WithContext(ctx).Limit(batchSize).Offset(offset).Find(&orgs).Error; err != nil {
+			return fmt.Errorf("unable to fetch orgs: %w", err)
+		}
+		if len(orgs) == 0 {
+			return nil
+		}
+
+		for _, org := range orgs {
+			if err := authz.ReconcileOrgRoles(ctx, db, org); err != nil {
```
ALPHAvibe, line 30: would this need to iterate over several records to reconcile org roles? Probablly worth assessing against prod db. Normally heavy requests we move to an admin endpoint since they are not suited for a db migration script.

jordan-acosta: It would, although the queries are cheap. But I'll refactor it anyways. This would become more expensive if we add more roles.

jordan-acosta: I optimized `ReconcileOrgRoles` to reduce the number of queries it was making, but digging into this turned up a bigger issue: the roles table wasn't index by org. So no only was I making too many `SELECTS`s, but they were much slower than I thought.

I update the PR to add an index. This is especially important since the dashboard is now reading org roles from the API.

jordan-acosta: I can still move this to endpoint if that's the SOP for backfills like, but I figured I should make this function more performant regardless of how we run it.

ALPHAvibe (approved):

PR #2133, opened 2026-08-10
chore: use shallow clone for potential branches

jonmorehouse:
Adds logic to attempt a shallow clone when a target ref is not a full commit sha.


RealHarshThakur (approved):

pkg/workspace/git.go
```diff
@@ -15,15 +15,99 @@ import (
...
+			URL:           w.src.URL,
+			ReferenceName: refName,
+			SingleBranch:  true,
+			Depth:         1,
+			Tags:          git.NoTags,
+			Progress:      pWriter,
+		})
+		if err == nil {
+			return nil
+		}
```
someshkoli, line 88: can cause issue with not iterating over entire list if first one succeeds

PR #2143, opened 2026-08-10
fix: slow get list service account

ALPHAvibe:
## fix: slow GET /v1/service-accounts

`GET /v1/service-accounts` was p95 4.5s and max 43.4s in prod, tripping the
`ui.request.latency` monitor on `endpoint:/v1/service_accounts`.

The whole cost was one query. In the 43s trace, 13 of the 14 Postgres spans were
1-20ms and one `query accounts` span was 43,356ms. The same `query accounts`
resource on every other route is p95 5-43ms, so this was the query shape, not
the database.

`ORDER BY accounts.email` plus `LIMIT` let the planner walk `idx_accounts_email`
in sort order and nested-loop into `account_roles` per row, betting it would
fill the limit early. Almost no account in the table belongs to any one org, so
it walked nearly the whole table before it could stop. `GROUP BY accounts.id`
blocked the cheaper shapes. Warm cache that's ~240ms for 5 rows; cold it's tens
of seconds, which matches the "slow the first time, fine after" reports.

### DIFF
--- a/services/ctl-api/internal/app/accounts/service/service_account.go
+++ b/services/ctl-api/internal/app/accounts/service/service_account.go
@@ -110,6 +110,30 @@ func (s *service) getOrgServiceAccount(ctx context.Context, orgID, accountID str
 	return acct, nil
 }
 
+// orgServiceAccountIDs resolves the org's service accounts off the
+// account_roles org index. Joining accounts to account_roles in the list query
+// instead lets the planner walk the whole accounts table in email order to
+// satisfy the ORDER BY and LIMIT, which took tens of seconds on a cold cache.
+func (s *service) orgServiceAccountIDs(ctx context.Context, orgID string, includeRunners bool) ([]string, error) {
+	tx := s.db.WithContext(ctx).
+		Model(&app.AccountRole{}).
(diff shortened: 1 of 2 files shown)


services/ctl-api/internal/app/accounts/service/service_account.go
```diff
@@ -110,6 +110,30 @@ func (s *service) getOrgServiceAccount(ctx context.Context, orgID, accountID str
...
+		Where(app.AccountRole{OrgID: generics.NewNullString(orgID)})
+
+	if !includeRunners {
+		tx = tx.
+			Joins("JOIN roles ON roles.id = account_roles.role_id AND roles.deleted_at = 0").
+			Where("roles.role_type != ?", app.RoleTypeRunner)
+	}
+
+	accountIDs := []string{}
+	if err := tx.Distinct().Pluck("account_roles.account_id", &accountIDs).Error; err != nil {
```
erickyellott, line 130: what the heck does Pluck do

erickyellott (approved):

services/ctl-api/internal/app/accounts/service/service_account.go
ALPHAvibe: its a gorm helper to make a slice over the results for a single column

PR #2188, opened 2026-08-13
feat(slack): mention workflow triggerer and approval responder

someshkoli:
Slack lifecycle messages currently render the raw email of the person who triggered a workflow and other things.  This resolves those emails to workspace users via `users.lookupByEmail` and renders `<@user>` mentions, falling back to the plain email when the lookup misses.


someshkoli:
For later

PR #2201, opened 2026-08-14
fix: remove infinite temporal activity retries from some places

someshkoli:
Temporal codegen is missing configs which should eventually kill off an activity and not let it run indefinitely. 
Along with that this pr also adds some activity run boundaries to some of the most failed activities.


jonmorehouse (dismissed):

fidiego (approved):
LGTM

PR #2208, opened 2026-08-18
chore: remove lfs check temporarily

someshkoli:


fidiego (approved):
LGTM
trusting you on this one
dare i say we should pin to a commit

PR #2213, opened 2026-08-18
fix: include runner_init_script_url in aws install tfvars

jordan-acosta:
## Summary

This was missing from the tfvars for aws installs. Causing the runner to fall back to the tag `cloud`. Since `cloud` doesn't exist, it then falls back to the latest version.


fidiego (approved):
LGTM

PR #2215, opened 2026-08-18
docs: rename app-branches-walkthrough title, re-arrange side bar

matt-zach-s:


sharkymark (approved):
looks good, thanks @matt-zach-s

PR #2222, opened 2026-08-19
fix: group runner health checks by install

RealHarshThakur:
## Summary

Runner health-check metrics now include the same `install_name` tag used by heartbeat and process-start metrics, allowing per-install monitors to group consistently.

## What stays the same

A failed install-name lookup does not block the health check or suppress its metric.


fidiego (approved):
LGTM

PR #2226, opened 2026-08-19
fix: route syncer helper writes through the sync transaction

im-Amitto:
#2186 wrapped the app config sync in one transaction, but `CreateAppBranchConfig`, `GetComponentIDs` and the component dependency helpers still used their own `*gorm.DB`, so they could not see rows the sync had just created. Adding a new app branch failed with `violates foreign key constraint` and adding a component with dependencies failed with `some components not created yet`; both now take the transaction, matching the existing `WithDB` helpers.

Also bounds the sync activity's retries. It was unlimited attempts over 24h, and since the app's sync queue runs one signal at a time, a failed sync held the only in-flight slot for a day and every later sync for that app hung instead of erroring.

Verified against the dev stack with a real config (branch with a public repo, a component with a dependency, an action): reproduced both errors on main, both sync clean on this branch, and a failed sync now releases the queue in ~8s so the next sync succeeds. The syncer test harness now runs the sync inside a transaction, which is why these were invisible to it before.


fidiego (approved):
LGTM

PR #2248, opened 2026-08-20
fix: skip policy evaluation when none apply

onprem:
Treat a missing policies configuration as an expected no-policy result while preserving failures for other database errors.

Filter policies before loading approval plan data so components without applicable policies bypass evaluation cleanly instead of failing on unnecessary plan preparation.


fidiego (dismissed):
LGTM

fidiego (approved):
LGTM

PR #2256, opened 2026-08-21
feat(cli): run ad hoc install actions

RealHarshThakur:
## Summary

Run one-time commands or scripts against an install from the CLI, with optional raw output streaming until completion.

**Run an ad hoc command or local script.** The command uses the install selected with `nuon installs select`, with `--install-id` available as an override.

**Configure the action.** Set environment variables directly or from a dotenv file, and control the timeout, name, IAM role, and Kubernetes configuration.

**Wait for command output.** `--wait` follows the same command-output log contract as notebook cells, drains delayed logs, and exits nonzero for unsuccessful runs. JSON and agent output keep structured results on stdout and send logs to stderr.

## What stays the same

Without `--wait`, the command returns queued action metadata immediately. Existing action workflow commands are unchanged.

### DIFF
--- a/bins/cli/cmd/actions.go
+++ b/bins/cli/cmd/actions.go
@@ -1,7 +1,11 @@
 package cmd
 
 import (
+	"time"
+
 	"github.com/spf13/cobra"
+
+	"github.com/nuonco/nuon/bins/cli/internal/services/actions"
 )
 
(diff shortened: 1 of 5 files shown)


fidiego (approved):
hmm, this might make things too efficient for our 🤖 friends
ship it asap

PR #2274, opened 2026-08-24
feat: tag runner process metrics for offline alerting

RealHarshThakur:
## Summary

Runner process metrics now carry enough tags to alert on real runner deaths and watch version rollouts, without paying for a new metric series on every routine restart.

**Alert on unexplained runner deaths by install.** The process stop metric is now tagged `reason:offline` in both paths that detect a missed heartbeat. A monitor on `runner.process.stop{reason:offline}` grouped by install name fires faster than the heartbeat-absence monitor and names the impacted install directly. Future stop causes can be added as new reason values without breaking that monitor.

**Watch a version rollout sweep the fleet.** The process start metric now carries `runner_version`, so a dashboard can show old-version processes draining and new-version processes coming up during a promotion.

## What stays the same

Stop metrics were already only emitted on missed heartbeats; the new tag makes that explicit rather than changing when they fire. Per-process drill-down is unchanged: the process ID remains on trace spans.

## Mechanics worth flagging

The process ID tag was dropped from the start metric. Install processes restart every 8 hours and every promotion restarts the whole fleet, so each restart minted a brand-new metric series; removing the tag caps cardinality at one series per runner instead of one per restart.


fidiego (approved):
LGTM

PR #2290, opened 2026-08-25
feat: improved sign in page

matt-zach-s:


matt-zach-s:
The current sign in page lacks helper and explainer content. It just has "sign in with Google" 

This PR fixes that for people logging into app.nuon.co

nnnnat (approved):

PR #2302, opened 2026-08-26
fix: stop app deletes racing component cleanup

RealHarshThakur:
## Summary

App deletion now waits for queued component deletion workflows before removing the app. This prevents overlapping PostgreSQL deletes while preserving cascade cleanup for components that were not separately queued.

## What stays the same

Components without a queued deletion continue to be removed through the existing app cascade.


fidiego (approved):
LGTMc

PR #2303, opened 2026-08-26
fix(vcs): stop retrying removed GitHub installations

RealHarshThakur:
## Summary

Removed GitHub App installations now return a client error instead of a server failure. This prevents permanent connection failures from inflating endpoint errors and triggering trace alerts.

**Avoid retries for permanent connection failures.** Repository and branch requests stop retrying 4xx responses while retaining bounded retries for transient server failures.

**Distinguish removed installations from control-plane failures.** GitHub installation-token 404 responses remain 404 responses through the API rather than being reported as 500 errors.

## What stays the same

Healthy GitHub connections continue loading repositories and branches normally. The dashboard keeps its existing error presentation and does not guess whether a missing resource needs reconnection.

### DIFF
--- a/services/ctl-api/internal/app/vcs/helpers/client.go
+++ b/services/ctl-api/internal/app/vcs/helpers/client.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


fidiego (approved):
LGTM

PR #2309, opened 2026-08-26
feat(services/ctl-api): add an authenticated stacks namespace to the runner api

jordan-acosta:
## Summary

Re-applying #2272 (reverted in #2299). **1 of 4.**

This PR adds a new `stacks` service to the Runner API, designed to be used by the Stack SDK. It has two endpoints:

- `GET /v1/stacks/{install_id}/config`
- `POST /v1/stacks/{install_id}/phone-home`

These replace the public `GET /v1/stack-runs/{phone_home_id}/config`, and use token auth instead of relying on `phone_home_id` as an ersatz access token. The legacy phone home endpoint is untouched. Cloudformation, BIcep, and legacy GCP Terraform stacks will keep working as-is against the old endpoint.


erickyellott (dismissed):

services/ctl-api/internal/app/installs/service/service.go
```diff
@@ -362,11 +362,6 @@ func (s *service) RegisterInternalRoutes(api *gin.Engine) error {
 }
 
 func (s *service) RegisterRunnerRoutes(api *gin.Engine) error {
-	// Read-only, side-effect-free config fetch consumed by the Terraform
-	// provider's nuon_stack data source. Public: the per-stack-version
-	// phone_home_id in the URL path is the secret.
-	api.GET("/v1/stack-runs/:phone_home_id/config", s.GetInstallStackVersionConfig)
```
erickyellott, line 368: is this safe to remove?

services/ctl-api/cmd/gen/main.go
```diff
@@ -44,6 +44,8 @@ func generateRunnerSchema(ctx context.Context) error {
 		"--parseDependency",
 		"--parseInternal", "-g", "runner.go",
 		"--markdownFiles", "docs/runner/descriptions",
+		// stacks/runner is omitted: its only client is the hand-written sdks/stack,
```
jonmorehouse, line 47: Hmm, I think we still want this in the spec. Also, why doesn't the sdk use an API client? We don't use hand-written clients anywhere else, as far as I can tell.

services/ctl-api/internal/app/installs/helpers/build_installer_sdk_config.go
```diff
@@ -75,24 +81,42 @@ func (s *service) buildInstallerSDKConfig(ctx context.Context, installID string)
 		return nil, fmt.Errorf("render secrets config: %w", err)
 	}
 
-	// Customer install inputs — names only. Values come from the per-install
-	// inputs table at apply time, mirroring the TF tfvars contract which also
-	// writes `"name" = ""` and lets the runner read values at runtime.
+	// Real values, not names: the read is authenticated. Current value per
```
jonmorehouse, line 84: I'm not sure I follow this one. What's the purpose of this stack configuration? are we returning the current input values?

services/ctl-api/internal/app/installs/service/service.go
jordan-acosta: Yes. It's only used by the Stack SDK.

services/ctl-api/internal/app/installs/helpers/build_installer_sdk_config.go
jordan-acosta: Yes. The stack reads the install's current stack config -- including the customer-facing inputs and their current values -- from the control plane. It then layers whatever values the customer provides over them.

jonmorehouse (approved):

jonmorehouse (approved):

PR #2320, opened 2026-08-26
fix: tighten workflow and activity lifeline and workflow bug fixes

someshkoli:
This change set max activity retry to 3 unless retry policy is provided in the annotations, this is to prevent runaway retried, and only let it retry for transient failures. 
It also reduces the keep alive time for workflows which are parked and waiting on user inputs, although this is mostly fine but reducing it us allows us to think about new constructs around how we wanna handle such workflows.  

Bug fixes:

- [x]  Issue with workflow still showing running on dashboard, when it was cancelled during step generation
- [x] Dashboard still showing approval banner even after approval window expire
- [x] workflows now terminate gracefully after 3 days rather than living forever


erickyellott (dismissed):

jonmorehouse:
@someshkoli I feel like since #2281 landed, we should close this and instead, set up a set of reasonable label-based defaults for activities and start classifying them that way. Do you mind taking a pass at that, instead?

someshkoli:
> @someshkoli I feel like since #2281 landed, we should close this and instead, set up a set of reasonable label-based defaults for activities and start classifying them that way. Do you mind taking a pass at that, instead?

Makes sense, though I would still get this in and fast follow with label one.

onprem (approved):

PR #2387, opened 2026-09-03
fix(config): harden dir parsing and add source archive capture

RealHarshThakur:
## Summary

App config parsing is more robust, and every config sync now captures the authored source files alongside the parsed config. That capture is the groundwork for showing file-level diffs in the app-branch preview.

- **Every config sync keeps the files as authored.** When the control plane parses an app config, it stores the exact authored files next to the parsed result, so a later run can show what changed at the file level, not just in the parsed output.
- **File references that reach into parent directories now resolve.** Relative paths starting with ../ in file and template lookups were passed through untouched; they now resolve as local files.
- **Large configs are no longer cut off mid-parse.** The deadline for resolving file and template lookups was fixed; it now scales with the parse.
- Includes a Ladle exploration (run `bun run dev:ladle` in dashboard-ui, under Approvals/PlanDiffs) showing how the captured files would render as a source diff in the branch preview.

### DIFF
--- a/pkg/config/parse/dir/parse_file.go
+++ b/pkg/config/parse/dir/parse_file.go
@@ -1,7 +1,9 @@
 package dir
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -45,7 +47,7 @@ func (e ErrParseFile) Unwrap() error {
 	return e.Err
 }
 
-func (p *parser) parseFile(path string, obj any) (bool, error) {
+func (p *parser) parseFile(path, group string, obj any) (bool, error) {
 	if !strings.HasSuffix(path, p.opts.Ext) {
 		path = path + p.opts.Ext
 	}
@@ -66,15 +68,25 @@ func (p *parser) parseFile(path string, obj any) (bool, error) {
 		return false, nil
 	}
 
-	fh, err := p.fs.Open(path)
+	contents, err := p.fs.ReadFile(path)
 	if err != nil {
-		return false, errors.Wrap(err, "unable to open path")
+		return false, errors.Wrap(err, "unable to read path")
 	}
 
-	if err := p.opts.ParserFn(fh, path, obj); err != nil {
+	if err := p.opts.ParserFn(io.NopCloser(bytes.NewReader(contents)), path, obj); err != nil {
 		return false, ErrParseFile{
 			Err:  err,
 			Name: path,
+		}
+	}
+	if p.opts.OnParsedFile != nil {
+		if err := p.opts.OnParsedFile(ParsedFile{
+			Path:     path,
+			Group:    group,
+			Contents: contents,
+			Value:    obj,
+		}); err != nil {
+			return false, errors.Wrap(err, "unable to record parsed file")
 		}
 	}
 
--- a/pkg/config/parse/dir/parse_file.go
+++ b/pkg/config/parse/dir/parse_file.go
@@ -86,7 +86,7 @@ func (p *parser) parseFile(path, group string, obj any) (bool, error) {
 			Contents: contents,
 			Value:    obj,
 		}); err != nil {
-			return false, errors.Wrap(err, "unable to record parsed file")
+			return false, errors.Wrap(err, "onParse hook failed")
 		}
 	}
 
--- a/pkg/config/parse/get/get_all.go
+++ b/pkg/config/parse/get/get_all.go
@@ -184,6 +184,7 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string) 
 	prefixes := []string{
 		"http",
 		"./",
+		"../",
 		"git",
 		"file",
 	}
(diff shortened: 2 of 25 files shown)


pkg/config/parse/dir/parse_file.go
```diff
@@ -66,17 +68,27 @@ func (p *parser) parseFile(path string, obj any) (bool, error) {
...
 		}
 	}
+	if p.opts.OnParsedFile != nil {
+		if err := p.opts.OnParsedFile(ParsedFile{
+			Path:     path,
+			Group:    group,
+			Contents: contents,
+			Value:    obj,
+		}); err != nil {
+			return false, errors.Wrap(err, "unable to record parsed file")
```
someshkoli: misleading wrapped error, could be `onparse hook failed` or something

pkg/config/parse/get/get_all.go
```diff
@@ -263,6 +267,34 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string)
 	return string(content), nil
 }
 
+func (g *get) recordLocalFile(inputVal, pwd string) error {
```
someshkoli, line 270: get unit test coverage here

pkg/config/parse/get/get_all.go
```diff
@@ -263,6 +267,34 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string)
...
+	if g.opts.OnLocalFile == nil || (!strings.HasPrefix(inputVal, "./") && !strings.HasPrefix(inputVal, "../")) {
+		return nil
+	}
+
+	root, err := filepath.Abs(g.opts.RootDir)
+	if err != nil {
+		return errors.Wrap(err, "resolve config root")
```
someshkoli: unable to*

pkg/config/parse/get/get_all.go
```diff
@@ -263,6 +267,34 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string)
...
+	}
+	resolved, err := filepath.Abs(filepath.Join(pwd, inputVal))
+	if err != nil {
+		return errors.Wrap(err, "resolve local source file")
```
someshkoli: unable to*

pkg/config/parse/get/get_all.go
```diff
@@ -263,6 +267,34 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string)
...
+	if err != nil || relative == ".." || strings.HasPrefix(relative, ".."+string(filepath.Separator)) {
+		return nil
+	}
+	info, err := os.Lstat(resolved)
+	if err != nil || !info.Mode().IsRegular() || info.Mode()&os.ModeSymlink != 0 {
+		return nil
+	}
+	contents, err := os.ReadFile(resolved)
+	if err != nil {
+		return errors.Wrap(err, "read local source file")
```
someshkoli: same^

pkg/config/parse/get/get_all.go
```diff
@@ -263,6 +267,34 @@ func (g *get) processField(ctx context.Context, inputVal string, subdir string)
...
+	}
+	return g.opts.OnLocalFile(filepath.ToSlash(relative), contents)
```
someshkoli, line 295: this might store full path rather than root relative path

pkg/config/parse/parse_dir.go
```diff
@@ -133,7 +175,96 @@ func ParseDir(ctx context.Context, parseCfg ParseConfig) (*config.AppConfig, err
...
+		if kind != "" && name == "" {
+			source.recordError(fmt.Errorf("source file %s has no name field", file.Path))
+			return nil
+		}
+		source.recordError(source.archive.AddMember(kind, name, file.Path))
+		return nil
+	}
+}
+
+func localSourceFileRecorder(source *sourceCapture) func(string, []byte) error {
```
someshkoli, line 215: get unit test coverage

someshkoli (changes requested):

services/ctl-api/internal/app/apps/signals/branches/appconfig/execute.go
```diff
@@ -185,6 +185,14 @@ func (s *Signal) Execute(ctx workflow.Context) error {
 		closeLogStream()
 		return fmt.Errorf("unable to serialize intermediate config: %w", err)
 	}
+	var sourceConfigJSON []byte
+	if intermediateConfig.SourceArchive != nil {
+		sourceConfigJSON, err = json.Marshal(intermediateConfig.SourceArchive)
+		if err != nil {
```
ALPHAvibe, line 187: can we defer closing a stream here? just to make sure we always close it.

RealHarshThakur: Done — `closeLogStream` is now idempotent and invoked via `defer` immediately after the stream is created, so every return path (and any future one) closes it. The manual calls scattered through `Execute` are gone, and `syncAndFinalize` no longer takes the closer as a parameter. Fixed in ce9f67d9.

pkg/config/parse/get/get_all.go
RealHarshThakur: we have this- "relative, err := filepath.Rel(root, resolved)" just above.

someshkoli (approved):

PR #2399, opened 2026-09-04
feat(api): workflow summary endpoint

nnnnat:
Adds workflow plan summary endpoint


services/ctl-api/internal/app/installs/service/get_workflow_plan_summaries.go
```diff
@@ -0,0 +1,515 @@
...
+	ctx *gin.Context, orgID string, pending []countableApproval,
+) (map[string]string, error) {
+	ids := make([]string, 0, len(pending))
+	for _, item := range pending {
+		ids = append(ids, item.approval.ID)
+	}
+
+	var rows []app.WorkflowStepApproval
+	if err := s.db.WithContext(ctx).
+		Select("id", "contents").
```
ALPHAvibe, line 292: contents is deprecated. we should be using contentsBlob.
but loading a lot of records with contents is risky in memory. Some potential for OOM here.

nnnnat:
Going to revisit this with a better approach

PR #2410, opened 2026-09-08
feat: remove component deployment option from stack reprovision

erickyellott:
There has been some confusion when doing a stack only reprovision. By default we didn't deploy components in the workflow, which is the opposite of sandbox reprovisions. 

We assume most of the time when running a stack reprovision you only want to do the stack/runner, and if  you need to do components, you may as well do the whole install reprovision to get the sandbox as well which depends on the stack. 

Here we remove the option to deploy components with the stack, and stop after the runner is up and healthy.


fidiego (commented):
LGTM

onprem (approved):

PR #2466, opened 2026-09-11
feat: log successful image signature verification during builds

RealHarshThakur:
Image builds with `require_signature = true` now log when signature verification passes, so a successful check is visible in the build log instead of being inferred from the build reaching the copy step.


fidiego (approved):
LGTM

PR #2499, opened 2026-09-15
fix: feature parity for container actions

ALPHAvibe:
## Summary

Container actions previously required `inline_contents` on every step. They now
also accept `command`, `public_repo`, and `connected_repo`, matching host
actions. Each step still takes exactly one source.

The runner writes `command` to a script and clones repo steps into the workspace
mount so the supervisor can run them inside the image. Outputs via `nuon_output`
and `$NUON_ACTIONS_OUTPUT_FILEPATH` work for these steps as well.

Image-backed runs create two jobs, an oci-sync job that mirrors the image and an
actions-workflow job that runs the steps. We were attaching the sync job as
`runner_job`, so the UI and `nuon installs actions outputs` surfaced the mirrored
image instead of step outputs. The read path now loads the actions-workflow job.
Host actions only ever have the one job, so their behavior is unchanged.

Docs updated to also reflect updates from this pr.


erickyellott (approved):
🔥

PR #2503, opened 2026-09-15
fix: (ctl-api) prevent stale log streams from blocking queues

onprem:
## Changes
- Clear inherited log-stream context before queue startup so retries do not fail on incomplete persisted streams.
- Make URL validation introduced in #2460 warning-only; preserve existing exporter initialization errors and URL path normalization.


fidiego (approved):
LGTM

PR #2507, opened 2026-09-15
fix: bound component configs on base tables

im-Amitto:
Alternative to #2505. Keeps the v2 view and rewrites only the hot query.

**Problem.** Since #2483, the missing-component resolution in `GetFullAppConfig` filters and sorts on `app_config_version` from `component_config_connections_view_v2`. That column is a correlated `row_number()` subquery, so Postgres re-runs the window over every app config of the app for every candidate row. On the largest app (911 configs, 17k connections) that is 15,370 × 911 rows per call.

**Change.** Query the base tables and bound by the app config's `(created_at, id)`, which is exactly what `app_config_version` orders by. Uses `idx_component_config_connections_component_id` and `idx_app_configs_preload`.

**Measured on the prod replica, same app, same 31 components.** Result sets compared with `EXCEPT` both ways: identical.

| | before (v2 view) | after |
|---|---|---|
| time | 21.6 s | 30 ms |
| buffers | 12.0M | 3.9k |

**Why not revert to v1 (#2505).** v1 windows the whole table on every point lookup: `WHERE id = ?` is 407 ms / 81k buffers and runs on every `ComponentConfigConnection` fetch and preload. That query was the top consumer in the 16:00 UTC spike today, before the promotion. v2 makes that lookup ~40 ms; this PR fixes the one query v2 made worse.


im-Amitto:
Superseded by #2508 now that main is back on the v1 view.

PR #2512, opened 2026-09-16
feat: view name plugin override

ALPHAvibe:
## Summary

When adding new views we need a way to toggle to the new view and revert back to the old view.

This pr adds a GORM views-plugin override so reads can be pointed at a different relation without making code changes.

Environment variables map a full source name to a full destination name. The variable suffix is the source view name.

## What's covered

- The default `{table}_view_vN` rewrite
- Custom views that skip versioning via `WithOverrideTable`, including `component_config_connections_latest_configs_view`
- `TableOrViewName` and `CurrentViewName`, which use the same rewrite
- Replica ACL, which uses the same rewrite

## Notes

- Migrations still create the original view names. This only changes what is queried.
- The destination relation must already exist.


services/ctl-api/internal/pkg/db/plugins/views/name.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/nuonco/nuon/services/ctl-api/internal/pkg/db/plugins"
 )
 
+// RewriteName returns the relation GORM should query. Env overrides replace
+// a full source name with a full destination name.
```
jonmorehouse: Why is this needed, when we use `TableOrViewName` everywhere else?

services/ctl-api/internal/pkg/db/plugins/views/views.go
```diff
@@ -74,13 +104,81 @@ func (m *viewsPlugin) modelsToViewTables(db *gorm.DB) {
 			modelType = modelType.Elem()
 		}
 
-		tableName := db.NamingStrategy.TableName(modelType.Name())
+		tableName := plugins.TableName(db, model)
```
jonmorehouse: why does this logic need to change? We use the strategy to go from model -> name, why the pivot etc?

services/ctl-api/internal/pkg/db/plugins/views/views.go
```diff
@@ -74,13 +104,81 @@ func (m *viewsPlugin) modelsToViewTables(db *gorm.DB) {
...
+		return nil
+	}
+	p, ok := db.Config.Plugins[PluginName].(*viewsPlugin)
+	if !ok {
+		return nil
+	}
+	return p
+}
+
+func loadEnvNameOverrides() map[string]string {
```
jonmorehouse: Is there a reason we need to load this override up front? Fine to leave it, but it does mean we have to load + parse the entire env, which is going to be slow and possibly sensitive. 

I think a better solution would be to be deterministic. When we are overwriting the table, we should build the expected env-var and look it up, vs reading + parsing all env-vars.

PR #2513, opened 2026-09-16
fix: make cancellation durable

RealHarshThakur:
- **Reject user actions on a cancelled workflow.** Skip, retry, and approve on a cancelled flow fail fast instead of touching a workflow that already stopped. This fixes the one cancellation test that failed i.e  cancelled workflow still accepted a skip.
- **Cancel a flow and watch it stop.** Cancel and shutdown signals reach every running step handler, each acknowledges before the workflow closes, and nothing overwrites the cancelled status on the way out.
- **Fail fast on workflows with no steps.** They error with a clear message instead of wedging with a generic failure.


nnnnat (dismissed):

services/ctl-api/internal/pkg/flow/signals/executeflow/execute.go
```diff
@@ -358,7 +358,21 @@ func (s *Signal) handle(ctx workflow.Context, startFromGroupIdx int) error {
 		}
 
 		if flw.GenerateStepsSignal == nil || flw.GenerateStepsSignal.Signal == nil {
-			return errors.Errorf("workflow %s has no steps and no generate-steps signal", s.WorkflowID)
+			missingStepsErr := errors.Errorf("workflow %s has no steps and no generate-steps signal", s.WorkflowID)
```
someshkoli: This state should never be possible, ideally

services/ctl-api/internal/pkg/flow/signals/executeflow/execute.go
```diff
@@ -968,6 +982,16 @@ func (s *Signal) isWorkflowComplete(ctx workflow.Context) bool {
 			app.WorkflowStepApprovalStatusApproved,
 			app.WorkflowStepNoDrift, app.WorkflowStepDrifted:
 			continue
+		case app.StatusError:
```
someshkoli: Errors are not terminal iirc ? Deploy could error still can  be retired or skipped, same for actions

onprem (approved):

PR #2517, opened 2026-09-16
feat: treat named policies as stand alone property in app config and epxose them via stack outputs

someshkoli:
Named policies are currently bounded to permission roles, it can be further utilized to be used for roles created in sandbox for irsa or other related trusts or other similar use cases.

This change exposes named property in stack outputs, which can then be accessed via states into downstream contents


im-Amitto (approved):
No overlap with the queue-routing work. This is stack outputs / named-policy config diff only, so it does not need changes from that PR.

PR #2526, opened 2026-09-16
fix: sample prompts not queries

jordan-acosta:
## Summary

The MCP docs should read "Sample prompts" not "Sample queries".


docs/guides/agents/sample-queries.mdx
```diff
@@ -1,8 +1,8 @@
 ---
-title: "Sample queries"
```
jonmorehouse, line 2: The file name is not updated here, and since it needs to be updated, we need to add a redirect for places this has been shared.

RealHarshThakur (approved):

PR #2531, opened 2026-09-16
feat: bulk cancel workflow

ALPHAvibe:
## Summary

  Adds a "Bulk cancel workflows" page to the admin dash, to allow bulk cancelling workflows by filter. You can cancell all or bulk select which workflows to cancel.
  
  Job will run as a background temporal job under `general-bulk-cancel-workflows` signal.
  When the cancell is submitted the user will see the signals page in the admin view for the triggerred queue signal.


fidiego (approved):
LGTM

PR #2536, opened 2026-09-17
chore(dashboard-ui): modify role selection modals

fidiego:
### Description

Two modals required modification

1. Reprovision Install: we never use roles here so we remove the role selection modal
2. Deprovision Install: we really want to use the Deprovision role but in some cases, there are operation roles that should be preferred. We should render it so we can choose it.


ALPHAvibe:
https://github.com/nuonco/nuon/blame/1c8a7d398d83130cff52230d06af9558162c85cb/services/ctl-api/internal/pkg/operation-roles/default_role.go#L23

we need to update this default map as well

ALPHAvibe (approved):

nnnnat (approved):

PR #2537, opened 2026-09-17
chore: add platform filter to org installs list

fidiego:
### Description

Add a cloud platform filter to the org installs api and dashboard-ui view to make the view more ergonomic for orgs with a large number of installs.

image


services/dashboard-ui/client/components/installs/InstallPlatformFilter/InstallPlatformFilter.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { Dropdown } from '@/components/common/Dropdown'
+import { Icon } from '@/components/common/Icon'
+import { Menu } from '@/components/common/Menu'
+import { Text } from '@/components/common/Text'
+import { CheckboxInputWithButton } from '@/components/common/form/CheckboxInput'
+import type { TCloudPlatform } from '@/types'
+
+const FILTER_OPTIONS: TCloudPlatform[] = ['aws', 'azure', 'gcp', 'unknown']
+const PLATFORM_LABEL: Record<TCloudPlatform, string> = {
+  aws: 'AWS',
```
ALPHAvibe, line 14: is this hard coded like this in other places? i would assume we have generated enum for this since its in enum on the schema

services/ctl-api/internal/app/installs/service/cloud_platform_filter.go
```diff
@@ -0,0 +1,77 @@
...
+		selectedTypes = append(selectedTypes, app.RunnerTypesForCloudPlatform(platform)...)
+	}
+
+	if !includeUnknown && len(selectedTypes) == 0 {
+		return tx
+	}
+
+	installTable := views.TableOrViewName(db, &app.Install{}, "")
+	typeExpr := fmt.Sprintf(`COALESCE(
+		(SELECT type FROM app_runner_configs WHERE app_config_id = %[1]s.app_config_id AND deleted_at = 0 LIMIT 1),
```
ALPHAvibe, line 59: you can also use tablevieworname here for AppRunnerConfig

services/dashboard-ui/client/components/installs/InstallPlatformFilter/InstallPlatformFilter.tsx
fidiego: can't hardcode w/ anything from the spec. the ts spec doesn't emit runtime enums.

services/ctl-api/internal/app/installs/service/cloud_platform_filter.go
fidiego: tell me more?

PR #63, opened 2025-12-26
feat: config generator, aka nuon apps init

someshkoli:
this pr adds support for `nuon apps init` command which is basically a local scaffolding tool that can be used to generate app configs via cli.

Currently supports generating entire custom config, build from existing templates, and add elements to app config incrementally.

More info can be found in docs file.

Has some side changes:
- fixes jsonschema for components and removes duplicated field definitions
- updates lsp and docs generator to respect the same

PR #2539, opened 2026-09-17
chore: require cloudsql db password

im-Amitto:
GCP custom stacks now include a Cloud SQL module whose password is required at apply time. This rejects a cloudsql stack with an empty `parameters.db_password` at sync, the same way dns already requires `dns_name`, and documents the module in the custom nested stacks guide.


im-Amitto:
Superseded. Cloud SQL should take the password from the install db_password secret, not a stack.toml parameter. That lives in install-stacks #46 and terraform-gcp-stack #8; kitchen-sink #63 drops the hardcoded value.

PR #2545, opened 2026-09-17
fix: retry & skip tests in integration suite

RealHarshThakur:
**Retry a failed step after the workflow has errored.** The workflow resumes from that step's group. If the step's automatic retries were already used up, a manual retry still starts a fresh attempt.

**Skip a whole group from a failed run.** When the skipped step's signal is defined as skip-group, the remaining steps in that group are marked discarded and the workflow moves on, matching what happens when the same skip lands during a live retry countdown.

**Declare a signal non-skippable.** Skippability is now a property a signal declares, defaulting to skippable. Used in runner-health and install-stack-version signals as those can't be skippable. 

**Faster, clearer failures for actions that cannot proceed.** A retry or skip on a cancelled workflow, or on one that already finished, fails immediately with a message saying so instead of waiting on a timeout. Re-delivering a retry to a step that was already retried is a no-op, so a double-click cannot mark the step retried twice.

## Follow up

A skip on a failed component apply with post-deploy actions would allow to skip the entire group instead of running those post deploy actions against an unapplied component.


services/ctl-api/internal/pkg/flow/signals/executeflow/update_skip_step.go
```diff
@@ -19,15 +25,186 @@ type SkipStepResponse struct {
...
+
+	// resumeRequested must be set last. The paused Execute loop acts on this
+	// flag the instant it flips, and reads the fields written just above it.
+	// The DB lookup above pauses this handler long enough for Execute to run,
+	// so setting the flag first means starting the skip from a stale
+	// resumeStartIdx.
+	s.resumeRunType = app.WorkflowRunTypeSkip
+	s.resumeStepID = req.StepID
+	s.resumeStartIdx = s.findGroupPositionForStep(ctx, req.StepID)
+	s.resumeRequested = true
```
onprem: Instead of repeating this logic everywhere we should make a helper to `markResumeRequsted()`. As this is being done in multiple places.

onprem (commented):
LGTM for most of the things. Just one nit that we can address before.

onprem (approved):

jonmorehouse (approved):

PR #2547, opened 2026-09-17
refactor(ui/api): better composite error message & treatment for workflows

nnnnat:
## ctl-api

- Parked steps that get abandoned keep their cause: the abandon status now carries the original `composite_error`, `metadata.original_error`, and a `step abandoned after failure: <cause>` description instead of overwriting them.
- New `deployerrors.DeployPlanRenderError` (`deploy.plan_render_failed`, fatal, terminal hint) with why/detail/how-to-fix sections.
- All deploy-plan render failures (terraform, helm, kubernetes manifest, pulumi) return that typed non-retryable error instead of a bare wrap.
- New `SetInstallDeployPlanCompositeError` activity freezes it onto the `install_deploys` row; wired into the 5 deploy/teardown/helm-recover signals behind version gates, and cleared at the start of each plan attempt.
- `GetStepErrorHints` reads the deploy row's composite error before falling back to the latest runner job, so the terminal hint stops the workflow instead of burning retries.
- Error detection walks the whole Temporal error chain (`errors.As` alone returns the outermost `ApplicationError`, whose type is the raw template message), and the recorded detail is the innermost message rather than the full wrapper chain.
- The activity resolves the component name from `install_components → components` when the deploy row doesn't carry it.
- `FlowStoppedErr` carries the failing step's composite error; the workflow status write adds `composite_error`, `step_name`, and `stop_reason`, preferring the original error over the placeholder description.

## dashboard-ui

- Workflow stopped banner is a one-liner naming the failing step, linking to its detail panel; it no longer repeats text already in the status line.
- Failed-step banners moved above the workflow header so the error is the first thing on the page.
- Composite error renders once, in the step banner — the step owns it.
- Step banner shows the original error for abandoned steps and falls back to `metadata.reason` when there's no status description.
- New `useStepErrorLog` hook + `StepErrorLog` component surface the step's latest error-severity log when no composite error exists.
- Step banner wraps its output in a gapped flex column so stacked banners aren't flush.
- Stories for the new states (plan-render failed, abandoned with/without composite error, stopped workflow variants, `StepErrorLog`) and a `StepBanner` test for the abandoned case.


services/ctl-api/internal/app/installs/deployerrors/plan_render_error.go
```diff
@@ -0,0 +1,109 @@
...
+		why = fmt.Sprintf("The %s for %s referenced a value that could not be resolved, so no plan could be built. Nothing was applied to your cloud account.", e.Stage, name)
+	}
+
+	sections := []compositeerrors.Section{
+		compositeerrors.MarkdownSection("Why", why),
+	}
+	if e.Detail != "" {
+		sections = append(sections, compositeerrors.CodeSection("Error detail", e.Detail))
+	}
+	sections = append(sections, compositeerrors.MarkdownSection("How to fix", "Check the referenced value exists in your app config — a missing sandbox output, action workflow output, or input is the usual cause. Fix the reference, re-run `nuon apps sync`, then retry the deploy."))
```
onprem, line 55: Should we remove the explicit `nuon apps sync` reference here and instead just ask them to update the app config and roll it out to the install before retrying.

onprem (approved):
Looks good, just one comment (but not blocking).

PR #2562, opened 2026-09-18
fix: Layer queues so a signal never waits on its own queue

im-Amitto:
App-branch runs wedge at "generating steps". A queue slot is held for a signal's entire execute phase, including time parked waiting on a child, so a parent and the child it awaits can never share a queue: once parents fill max_in_flight, the run hangs until the 3-day timeout.

- Prod: of 3140 branch execute-workflow signals, only 557 were in the right queue. 410 went to app-branch-generate-steps and 460 to app-branch-workflow-step-groups, both 2-slot queues where it waits on its own children. Cause: GetQueueByOwner filtered on owner and called First(), ORDER BY id over random nanoids.
- One queuenames registry owns names, defaults and capacities. Every enqueue names its queue.
- Runtime guard rejects a callback-bearing enqueue into the caller's own queue, AST tests fail the build on unnamed enqueues. A silent multi-day stall becomes an immediate error naming the queue, visible on the step.
- Avoids the #2438 revert: each owner type declares its own default rather than pinning name = '' everywhere, since installs, runners and orgs have no unnamed queue.
- The guard already caught a live one: provision-sandbox-plan awaited workflow-step-approval-request on the same install-signals queue. Approvals moved to install-approvals.
- Migration soft-deletes duplicate (owner, name) rows (14 component pairs in prod) and adds a unique index.
- Verified live, not just tests: real org, branch-backed app, 9 components including 2 images, real CloudFormation stack and EKS sandbox, components deploying. Placement confirmed in the DB.
- Low risk: no worker or task-queue changes, old activity types stay as shims for one release, existing branches pick up the new layer on their next run, rollback is a plain revert.


services/ctl-api/internal/app/vcs/service/write_event.go
```diff
@@ -85,7 +85,7 @@ func (s *service) fanOutToVCSConnections(ctx context.Context, event *app.GithubE
 			continue
 		}
 
-		queue, err := s.queueClient.GetQueueByOwner(connCtx, conn.ID, "vcs_connections")
+		queue, err := s.queueClient.GetOnlyQueueByOwner(connCtx, conn.ID, "vcs_connections")
```
someshkoli: note to myself: check what this does

services/ctl-api/internal/pkg/config/syncer/stack/sync.go
```diff
@@ -42,7 +42,7 @@ func Sync(ctx context.Context, db *gorm.DB, appsHelpers *appshelpers.Helpers, cf
 
 	// Templates upload to S3 asynchronously; until then each stays pending and
 	// is skipped at stack generation.
-	q, err := appsHelpers.QueueClient().GetQueueByOwner(ctx, appID, "apps")
+	q, err := appsHelpers.QueueClient().GetDefaultQueueByOwner(ctx, appID, "apps")
```
someshkoli: self note: check what this does

someshkoli (approved):

PR #2569, opened 2026-09-18
ci: compact go cache on warm mounts

im-Amitto:
Go build cache compaction works, but it is skipped whenever the restore stage takes its warm-mount shortcut. The restore stage skips the S3 download when the cache mount already holds the seed we were about to fetch, which happens when a PR build leaves its cache on the same isolated runner that the follow-up main build lands on. The compaction marker was written inside that download branch, so on the warm path it was never created and the upload stage silently skipped compaction.

Run 35383271467 shows both behaviours side by side. cli, lsp and runner all downloaded their seeds and compacted: 4.7G to 2.6G, 806M to 513M, and 5.5G to 4.2G respectively, plus 8.5G to 4.7G for the cli release cache. ctl-api found its mount already current, so it compacted nothing and uploaded the full 29G.

This moves the marker setup out of the download branch so it runs whether or not we had to fetch the seed, across all seven restore stages (ctl-api, and the build and release stages of cli, lsp and runner).

Also confirmed the epoch stamp is load-bearing, using golang:1.25 since that is what the build image is based on. Stamping 443 cache entries to epoch and rebuilding left 314 with refreshed mtimes and 129 untouched; dropping those 129 still gave a fully cached rebuild with no compile invocations. Go only refreshes an entry's mtime when it is over an hour old, so without the stamp a freshly extracted cache would look entirely unused.


im-Amitto:
Not needed. The warm-mount path is rare (1 of the last 8 main runs) and self-corrects on the next run: ctl-api compacted 48G to 5.4G in run 35384176644, right after the 29G miss.

PR #2572, opened 2026-09-20
fix: runner status during provision

jordan-acosta:
## Summary

The "Await runner healthy" step has been failing intermittently during provision workflows.

The root cause in this case appears to be using both the status and status_v2 values in different places. This PR updates the runner status check to check both. It also adds polling to the runner status await to guard against any other delays or race conditions when provisioning a runner.

A follow may be needed to align the right status to update and check.


services/ctl-api/internal/app/installs/signals/awaitrunnerhealthy/signal.go
```diff
@@ -138,12 +146,33 @@ func (s *Signal) Execute(ctx workflow.Context) error {
...
 		InitialInterval: time.Second * 15,
 		MaxInterval:     time.Minute * 1,
 		BackoffFactor:   1.1,
 		Fn: func(ctx workflow.Context) error {
+			if failFast && debounce {
+				r, err := activities.AwaitGetRunnerByID(ctx, install.RunnerID)
+				if err != nil {
+					return err
+				}
+				runner = r
```
someshkoli: useless reassignment

services/ctl-api/internal/app/runners/worker/activities/healthcheck_decisions.go
```diff
@@ -67,7 +67,7 @@ type runnerHealthDecision struct {
 func decideRunnerHealth(now time.Time, runner *app.Runner, presence runnerProcessPresence) runnerHealthDecision {
 	var d runnerHealthDecision
 
-	if isSkippableRunnerStatus(runner.Status) {
+	if isSkippableRunnerStatus(runner.Status) || isSkippableRunnerStatus(app.RunnerStatus(runner.StatusV2.Status)) {
```
someshkoli: a method on runner called Status() would go long way to get rid of older status usage and prevent similar problems

services/ctl-api/internal/app/runner.go
```diff
@@ -41,9 +41,12 @@ const RunnerOfflineTSMetadataKey = "offline_ts"
 // Boot window after a stack apply before a missing process counts as offline.
 const RunnerAwaitingHeartbeatGrace = 30 * time.Minute
 
-// HealthcheckPending is true while a missing process is expected (stack not applied yet, or still booting).
 func (r *Runner) HealthcheckPending(now time.Time) bool {
-	switch r.Status {
+	return r.statusPending(r.Status, now) || r.statusPending(RunnerStatus(r.StatusV2.Status), now)
```
someshkoli: same here too ^, avoids unnessary bloat for such backward compatibility

im-Amitto:
Superseded by #2577, which separates startup readiness from day-2 fail-fast checks and makes runner status transitions atomic.

PR #2585, opened 2026-09-21
chore: remove component-health feature flag

im-Amitto:
Drops the `component-health` org flag so the feature is on for every org. Removes the gates from the health read endpoints, the runner ingest/context endpoints, the evaluator activity, the verified-deploy gate step, and the dashboard Resources tab / Health column. No data migration: `Org.AfterQuery` prunes the stale key.


erickyellott:
Do we need to start any emitters or run any scripts for existing orgs?

erickyellott (dismissed):

erickyellott (approved):

